PR #17, opened 2020-03-21
Add last_scan to RescanRunner

robbevp:


app/models/rescan_runner.rb
```diff
@@ -41,7 +42,7 @@ def run
     rescue StandardError => e
       update(error_text: "#{error_text}A really unexpected error occurred while processing: #{e.message}\n#{e.backtrace.join("\n")}\n")
     ensure
-      update(running: false)
+      update(running: false, last_scan: DateTime.current)
```
charvp: I think `DateTime.now` is more common (and clear) than `current`?

robbevp: `Time.now` could load to issues with timezones: https://stackoverflow.com/a/18811305

charvp: Right, forgot about that.

PR #18, opened 2020-03-24
Don't add 'Various artists' as albumartist

### LINKED ISSUE #8: Try to detect various artists in rescan runner

robbevp:
fix #8

### DIFF
--- a/app/models/rescan_runner.rb
+++ b/app/models/rescan_runner.rb
@@ -94,16 +94,18 @@ class RescanRunner < ApplicationRecord
         next
       end
 
-      albumartist = [AlbumArtist.new(artist: (Artist.find_by(name: t_albumartist) || Artist.new(name: t_albumartist, review_comment: 'New artist')), 
-                                     name: t_albumartist, 
-                                     order: 1, 
-                                     separator: nil)] unless t_albumartist.downcase == 'various artists'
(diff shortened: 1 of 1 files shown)


app/models/rescan_runner.rb
```diff
@@ -96,12 +96,16 @@ def process_file(location, codec, path)
 
       albumartist = Artist.find_by(name: t_albumartist) || Artist.new(name: t_albumartist, review_comment: 'New artist')
```
charvp: As discussed, move the Various Artists if to here.

app/models/rescan_runner.rb
```diff
@@ -95,13 +95,21 @@ def process_file(location, codec, path)
...
+      albumartists = if t_albumartist.downcase != 'various artists'
+                       [AlbumArtist.new(artist: albumartist,
+                                        name: t_albumartist,
+                                        order: 1,
+                                        separator: nil)]
+                     else
+                       []
+                     end
+
+      album = Album.find_by(title: t_album, release: Date.ordinal(t_year)) || Album.new(title: t_album,
```
charvp: Is there a reason you moved the `Album.new` to this line? Because it's a lot more readable on the next, IMO.

robbevp: Have corrected this. Think this was rubocop's doing

PR #23, opened 2020-03-29
Use AMS as a JSON serializer

charvp:
Another performance improvement (although less drastic than #22). This (more or less) halves the time spent serializing. Total request time for albums (the slowest route) is reduced by about 25% (with #22 already applied).

In addition to the performance improvements, I feel like the code is a bit more readable. The logs on the other hand, become a *lot* more readable.


app/serializers/full_auth_token_serializer.rb
```diff
@@ -0,0 +1,3 @@
+class FullAuthTokenSerializer < AuthTokenSerializer
```
robbevp: I find this naming a bit odd - it would be easier to understand if this was named `AuthToken...Serializer`

PR #28, opened 2020-04-04
Allow merging of genres

robbevp:


app/controllers/genres_controller.rb
```diff
@@ -42,6 +42,10 @@ def destroy_empty
     Genre.left_outer_joins(:tracks).where(tracks: { id: nil }).destroy_all
   end
 
+  def merge
+    render json: @genre.errors, status: :unprocessable_entity unless @genre.merge(Genre.find(params[:old_genre_id]))
```
charvp, line 46: The error handling here doesn't make much sense. `@genre` is never edited so will never get errors and if `@genre.merge(...)` returns something falsy it will be because the `other.destroy` statement failed.

app/controllers/genres_controller.rb
```diff
@@ -42,6 +42,10 @@ def destroy_empty
     Genre.left_outer_joins(:tracks).where(tracks: { id: nil }).destroy_all
   end
 
+  def merge
+    render json: @genre.errors, status: :unprocessable_entity unless @genre.merge(Genre.find(params[:other_genre_id]))
```
charvp, line 46: I think we should just say (like for tracks) that the `merge` method in `Genre` shouldn't be able to fail (and test that extensively). Because now this action works in the opposite way as it does in the tracks controller, which doesn't seem very consistent.

Rendering the updated genre (like in the tracks controller) would also be good. Just for consistency of course, since the model won't actually be changed.

app/models/genre.rb
```diff
@@ -15,4 +15,12 @@ class Genre < ApplicationRecord
   validates :name, presence: true, uniqueness: true
 
   normalized_col_generator :name
+
+  def merge(other)
+    tracks = self.tracks
+    tracks.find_each do |track|
```
charvp, line 21: Why update via tracks? Can't you just add all tracks of the other genre to this genre? I think/hope that would make for a lot less database calls.

charvp: Ignore this, it doesn't seem like significantly reducing the number of database calls is possible, and using the `find_each` over tracks will result in less memory usage. I do worry that the action in the controller might take a very long time though, if a genre with a **lot** of tracks is merged.

app/models/genre.rb
```diff
@@ -15,4 +15,12 @@ class Genre < ApplicationRecord
   validates :name, presence: true, uniqueness: true
 
   normalized_col_generator :name
+
+  def merge(other)
+    tracks = other.tracks
+    tracks.find_each do |track|
```
charvp, line 21:
```suggestion
    other.tracks.find_each do |track|
```
You should avoid shadowing a method name.

app/controllers/genres_controller.rb
robbevp: I've made the action consistent with the way this is done in tracks

app/controllers/genres_controller.rb
```diff
@@ -42,6 +42,12 @@ def destroy_empty
     Genre.left_outer_joins(:tracks).where(tracks: { id: nil }).destroy_all
   end
 
+  def merge
+    @genre.merge(Genre.find(params[:other_genre_id]))
+    # We don't do error handling here. The merge action isn't supposed to fail.
+    render json: @genre, status: :created
```
charvp, line 48:
```suggestion
    render json: @genre, status: :ok
```

test/models/genre_test.rb
```diff
@@ -16,4 +16,31 @@ class GenreTest < ActiveSupport::TestCase
     assert_not genre.normalized_name.nil?
     assert_equal 'iouaaa', genre.normalized_name
   end
+
+  test 'should be able to merge genres' do
+    genre1 = create(:genre)
+    genre2 = create(:genre)
+
+    assert genre2.merge(genre1)
```
charvp, line 24: If the assumption is that `merge` shouldn't fail, `assert`ing it's return value doesn't make much sense.

PR #29, opened 2020-04-04
Use new serializing for track merge action

charvp:
Also improve tests, because this should have been caught when doing the change. Also do some better testing that the merge action in the track model can't fail.

### DIFF
--- a/test/models/track_test.rb
+++ b/test/models/track_test.rb
@@ -16,13 +16,52 @@
 require 'test_helper'
 
 class TrackTest < ActiveSupport::TestCase
-  test 'should be able to merge two tracks' do
+  test 'should be able to merge two tracks with two starting audio_files' do
     a1 = create(:track, :with_audio_file)
     a2 = create(:track, :with_audio_file)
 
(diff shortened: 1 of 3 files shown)


test/models/track_test.rb
```diff
@@ -25,6 +25,45 @@ class TrackTest < ActiveSupport::TestCase
...
+    a1 = create(:track)
+    a2 = create(:track, :with_audio_file)
+
+    a1.merge(a2)
+
+    assert_equal 1, AudioFile.count
+    assert a1.reload.audio_file.present?
+  end
+
+  test 'should be able to merge two tracks with zero resulting audio_files' do
```
robbevp, line 38: I don't see how this is the behaviour we want. Why would we want to end up with a track with audio?

charvp: I just added tests documenting the current behaviour of the method. If we want to change that behaviour (and I agree that the behaviour could be better) that should happen in a different PR.

robbevp: 👍

PR #31, opened 2020-04-04
Change track merge behaviour to avoid throwing away audio files

charvp:
Also improves tests some more.


charvp:
Had to push another change to appease the linter, can you approve again?

PR #33, opened 2020-04-04
Test the rescan runner

charvp:


charvp:
> I think we need a few more tests as a basis.
> At least one for a file without the necessary tags & one with various artists as album artists

That's why this is a draft PR and I didn't ask you to review yet. :wink:

PR #37, opened 2020-04-04
Allow artists to be merged

robbevp:


charvp:
I'm not in favor of allowing artist merge. 
* There are validations for `track_artists` and `album_artists` that won't be checked (or at least, can fail).
* A mass edit basically removes the need for this.

robbevp:
> I'm not in favor of allowing artist merge.
> 
>     * There are validations for `track_artists` and `album_artists` that won't be checked (or at least, can fail).
I don't see which validations are currently not covered?

>     * A mass edit basically removes the need for this.
This is true for albums, but not for artists. For example: If both the artists belong to tracks with many track_artists that differ from track to track, merging them in mass edit would be nearly impossible & a lot of work

charvp:
> I don't see which validations are currently not covered?

Order will not be set correctly (for both track artists and album artists) and separator will not be checked for album artists. (Especially relevant in the case that both artists are on the track/album.)

> This is true for albums, but not for artists. For example: If both the artists belong to tracks with many track_artists that differ from track to track, merging them in mass edit would be nearly impossible & a lot of work

I guess this is true, but I feel like this is a case that will not actually happen? The only realistic time I would see this being useful is when the rescan runner created an artist with a slightly wrong name. But in that case, the tracks will have at most 2 artists (main and composer). A mass edit in that case would be very easy.

robbevp:
IMHO the amount of times this will save the user time is bigger than the amount of cases where this could cause trouble.

> Order will not be set correctly (for both track artists and album artists) and separator will not be checked for album artists. (Especially relevant in the case that both artists are on the track/album.)

Having both of the on one track or album seems very rare to me, but then still only the separator seems like a real concern. The order would not throw any validation errors and would just be resolved the next time either the track or album is updated.
The validation of the separator would not fail until a user tries to update the album, since the `album_artist` is destroyed via the artist and the album is not touched, thus not triggering those validations to run.

We could either:
1. Add a `review comment` to those tracks/albums where the app removed an `album_artist`/`track_artist`. This might cause some confusion but would work, since we would add the review comment before the artist is destroyed so the separator validation would not yet throw an error.

2. Not change the `album_artist` and `track_artist` if they would be removed, not the delete the artist and return an error to the user saying that they have to resolve this case manually.

> I guess this is true, but I feel like this is a case that will not actually happen? The only realistic time I would see this being useful is when the rescan runner created an artist with a slightly wrong name. But in that case, the tracks will have at most 2 artists (main and composer). A mass edit in that case would be very easy.

The amount of work involved in this quickly adds up. (Certainly when there are different composers etc.)

Especially when users are adopting the app or still working through a backlog of review comments, this could save a lot of time. A user might not always know that there are two entries for the artist and only discover this after they have already updated some albums/tracks for both of the entries. (Speaking from experience here 😉)

robbevp:
> Did we get to a consensus on what we want to do here? Looking again at it now, maybe the solution is actually do error handling and make it possible for the action to fail? We could then handle the easy cases where the automation is possible, and where not, deliver a message to the user that they will have to do the work manually.

I agree with this solution. (And still feel very strongly that we should offer the user an easy option).
I'll rework this sometime over the next weeks or so

robbevp:
@charvp I've reworked this so we return an error if there is an overlap in tracks (with same role) or albums

PR #43, opened 2020-04-20
Allow a blank string as AlbumArtist separator

### LINKED ISSUE #42: AlbumArtists separator should allow single space

robbevp:
fix #42

### DIFF
--- a/app/models/album.rb
+++ b/app/models/album.rb
@@ -70,12 +70,9 @@ class Album < ApplicationRecord
     album_artists.each do |aa|
       if aa.order == album_artists.to_a.count
         errors.add(:album_artists, 'aa-last-no-separator') unless aa.separator.nil?
-      # rubocop:disable Rails/Blank
-      # We don't want to check for blank? here, since we specifically want to allow a blank string
-      elsif aa.separator.nil? || aa.separator.empty?
+      elsif aa.separator.nil?
         errors.add(:album_artists, 'aa-separator')
       end
-      # rubocop:enable Rails/Blank
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/album.rb
```diff
@@ -70,9 +70,12 @@ def album_artist_separators
     album_artists.each do |aa|
       if aa.order == album_artists.to_a.count
         errors.add(:album_artists, 'aa-last-no-separator') unless aa.separator.nil?
-      elsif aa.separator.blank?
+      # rubocop:disable Rails/Blank
+      # We don't want to check for blank? here, since we specifically want to allow a blank string
+      elsif aa.separator.nil? || aa.separator.empty?
```
charvp, line 75: Maybe we should allow the empty string as well? I'm thinking about cases like [NehruvianDOOM](https://www.discogs.com/NehruvianDOOM-NehruvianDOOM/master/741082) where the "Nehruvian" part is "Bishop Nehru" and the "DOOM" part is "MF DOOM".

robbevp: True, I've updated the PR

app/models/album.rb
```diff
@@ -70,9 +70,12 @@ def album_artist_separators
     album_artists.each do |aa|
       if aa.order == album_artists.to_a.count
         errors.add(:album_artists, 'aa-last-no-separator') unless aa.separator.nil?
-      elsif aa.separator.blank?
+      # rubocop:disable Rails/Blank
```
charvp, line 73: The disabling of the rubocop check shouldn't be necessary anymore?

PR #44, opened 2020-04-21
Allow scope by id on main models

### LINKED ISSUE #39: Allow get of multiple items
For accentor/web#77 and after merges, we might want to allow one request for multiple objects. 
Possibly by passing an `ids` parameter with an array for index?

robbevp:
fix #39

### DIFF
--- a/app/controllers/albums_controller.rb
+++ b/app/controllers/albums_controller.rb
@@ -2,6 +2,7 @@ class AlbumsController < ApplicationController
   before_action :set_album, only: %i[show update destroy]
 
   has_scope :by_filter, as: 'filter'
+  has_scope :by_ids, as: 'ids', type: :array
   has_scope :by_label, as: 'label'
   has_scope :by_labels, as: 'labels', type: :array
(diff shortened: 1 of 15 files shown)


charvp:
I still feel like this is something we don't really want to do. Do you still feel strongly about this feature? If not, you can close this (and the linked issue). Otherwise, we'll talk about this in-person some more.

You could also repurpose the linked issue into something else, namely that we do need a more coherent story on synchronization issues (but I fear that there is no easy solution there).

robbevp:
> I still feel like this is something we don't really want to do. Do you still feel strongly about this feature? If not, you can close this (and the linked issue). Otherwise, we'll talk about this in-person some more.
> 
> You could also repurpose the linked issue into something else, namely that we do need a more coherent story on synchronization issues (but I fear that there is no easy solution there).

I don't feel that strongly about it anymore - I think we found a good way to handle merges and deletions.
The synchronization will need some work/upgrades in the future, so I'll rework the issue

PR #51, opened 2020-05-19
Expose filename of audio_file

robbevp:
Will be necessary for Accentor/web#33

### DIFF
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -34,11 +34,7 @@ class TracksController < ApplicationController
     @track = Track.new(transformed_attributes)
 
     if @track.save
-      if current_user.moderator?
-        render json: @track, serializer: TrackModeratorSerializer, status: :created
-      else
-        render json: @track, status: :created
-      end
+      render json: @track, serializer: TrackModeratorSerializer, status: :created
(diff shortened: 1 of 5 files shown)


charvp:
I feel like this information should only be available to moderators. It exposes the underlying file structure to unprivileged users right now.

robbevp:
> I feel like this information should only be available to moderators. It exposes the underlying file structure to unprivileged users right now.

I've updated the MR. I suppose we will rename the new serializer if/when we want to expose other properties to mod+ only - but would only do that then

charvp:
> I've updated the PR. I suppose we will rename the new serializer if/when we want to expose other properties to mod+ only - but would only do that then

I would rename the serializer right now. Something like ModeratorTrackSerializer or similar. It makes the purpose more clear, IMO. Also, could you add some tests? Doesn't seem to hard, and I would like it if we raised the project coverage with every PR. Coverage isn't everything, but if we add meaningful tests every time, it should mean something.

robbevp:
@charvp There is only a small difference in coverage, but at least all changes are covered.
I've renamed the serializer `TrackModerator` - not the best name, but it had the benefit that the file was sorted alongside the track serializer

PR #68, opened 2020-07-04
Use wahwah to read audio metadata

### LINKED ISSUE #66: Experiment with wahwah for reading audio metadata
The current library, easytag, is unmaintained, so [wahwah](https://github.com/aidewoode/wahwah) might be a good alternative.

robbevp:
fix #66


robbevp:
@charvp WahWah doens't seem to be able to read the bitrate of an MP3 file with VBR and a duration < 1s. (It uses the duration to calculate the bitrate). 
Do you know if we have any files with such a short duration in production?

charvp:
> @charvp WahWah doens't seem to be able to read the bitrate of an MP3 file with VBR and a duration < 1s. (It uses the duration to calculate the bitrate).
> Do you know if we have any files with such a short duration in production?

The shortest track we have in production is 3 seconds long.

robbevp:
> > @charvp WahWah doens't seem to be able to read the bitrate of an MP3 file with VBR and a duration < 1s. (It uses the duration to calculate the bitrate).
> > Do you know if we have any files with such a short duration in production?
> 
> The shortest track we have in production is 3 seconds long.

Okay to update the testfiles to be >= 1s and provide a check so that a file that where WahWah can't determine a bitrate will give an error in the rescan?

charvp:
If I'm reading the WahWah source right, the bitrate will just be `nil`, right? Can't we just provide a default value ourselves? The bitrate isn't very useful for a track with length 0 anyway, so just setting it to 0 seems fine to me. Then we don't have to add larger binary files to the repo and we're future-proofed for if such a short track ever makes it into production.

robbevp:
> If I'm reading the WahWah source right, the bitrate will just be `nil`, right? Can't we just provide a default value ourselves? The bitrate isn't very useful for a track with length 0 anyway, so just setting it to 0 seems fine to me. Then we don't have to add larger binary files to the repo and we're future-proofed for if such a short track ever makes it into production.

Sure, that works as well. The bitrate wouldn't be very interesting either way. Because of the length of the file different applications seem to handle this very differently.

.github/workflows/ci.yml
```diff
@@ -26,7 +26,6 @@ jobs:
     - name: Install dependencies
       run: |
         apt-get -y update
```
charvp, line 28: This can be removed as well if we don't need taglib.

app/models/rescan_runner.rb
```diff
@@ -77,77 +77,76 @@ def process_file(location, codec, path)
...
+    t_number = tag.track || 0
+    t_album = tag.album&.unicode_normalize
+    t_year = tag.year.to_i
+    t_genre = tag.genre&.unicode_normalize
+    length = tag.duration
+    bitrate = tag.bitrate || 0
+
+    unless t_artist.present? && t_title.present? && t_album.present?
+      update(error_text: "#{error_text}File #{path} is missing required tags (album, artist, title)\n")
+      return false
```
charvp, line 94: This method doesn't return a value, so this should be a simple `return`.

charvp:
This section should also be removed from the dependabot config file: https://github.com/accentor/api/blob/develop/.github/dependabot.yml#L18

PR #74, opened 2020-07-06
Add sample rate and bit depth to audio file

### LINKED ISSUE #52: Add bit depth and sample rate to audio file
Will be relevant for Accentor/web#33

robbevp:
fix #52

### DIFF
--- a/db/migrate/20200705102418_add_sample_rate_and_bit_depth_to_audio_file.rb
+++ b/db/migrate/20200705102418_add_sample_rate_and_bit_depth_to_audio_file.rb
@@ -3,12 +3,9 @@ class AddSampleRateAndBitDepthToAudioFile < ActiveRecord::Migration[6.0]
     add_column :audio_files, :sample_rate, :integer
     add_column :audio_files, :bit_depth, :integer
 
+    AudioFile.update_all(sample_rate: 0, bit_depth: 0)
     AudioFile.find_each do |af|
-      next unless af.check_self
-      tag = WahWah.open(af.full_path)
-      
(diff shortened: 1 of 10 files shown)


db/migrate/20200705102418_add_sample_rate_and_bit_depth_to_audio_file.rb
```diff
@@ -0,0 +1,22 @@
+class AddSampleRateAndBitDepthToAudioFile < ActiveRecord::Migration[6.0]
+  def up
+    add_column :audio_files, :sample_rate, :integer
+    add_column :audio_files, :bit_depth, :integer
+
+    AudioFile.find_each do |af|
```
charvp, line 6: This migration would take hours, if not days. That's not acceptable. If we want this (which I'm still not convinced about), a job will need to be scheduled for each audio_file, so that the app is usable while these values are calculated.

charvp:
Also, I'm not a fan of depending on an unstable version of a library unless absolutely necessary. Can't we just wait until a new version is released?

db/migrate/20200705102418_add_sample_rate_and_bit_depth_to_audio_file.rb
robbevp: It should take about 20 to 30 minutes if I calculated it right.
But even then your comment is still true, I've updated the PR

robbevp:
> Also, I'm not a fan of depending on an unstable version of a library unless absolutely necessary. Can't we just wait until a new version is released?

Sure, we can keep this until a new version is released.

PR #77, opened 2020-07-07
Add rubocop-performance to rubocop config

robbevp:


charvp:
@robbevp Do you still want this? As mentioned IRL, I'm not a fan of a set of predetermined rules for performance (especially if it makes the code less clear).

robbevp:
No, I’m fine with dropping this

PR #82, opened 2020-07-20
Always allow byte-range requests

charvp:


app/jobs/transcode_cache_clean_job.rb
```diff
@@ -0,0 +1,7 @@
+class TranscodeCacheCleanJob < ApplicationJob
+  queue_as :transcode_cache_cleaner
+
+  def perform
+    TranscodedItem.where.not(last_used: (3.days.ago..)).destroy_all
```
robbevp, line 5: Isn't three days rather long? Since we only run this every 24h, this means that a file can exist up to 96h.

config/initializers/delayed_job.rb
```diff
@@ -3,3 +3,9 @@
 Delayed::Worker.max_run_time = 1.day
 Delayed::Worker.raise_signal_exceptions = :term
 Delayed::Worker.logger = Logger.new(Rails.root.join("log/delayed_job_#{Rails.env}.log"))
+Delayed::Worker.queue_attributes = {
+  transcodes: { priority: 5 },
```
robbevp, line 7: Shouldn't there be a queue for `AudioFile.check_file_attributes` ? It is meant to be ran as a delayed job

db/migrate/20200720092916_add_cron_to_delayed_jobs.rb
```diff
@@ -0,0 +1,7 @@
+class AddCronToDelayedJobs < ActiveRecord::Migration[6.0]
+  def change
+    add_column :delayed_jobs, :cron, :string
+
+    Delayed::Job.enqueue(TranscodeCacheCleanJob.new, cron: '0 0 * * *')
```
robbevp, line 5: Assuming the server is in a similar timezone to the user, I would shedule this a few hours later to get as little overlap with actual usage.

app/jobs/transcode_cache_clean_job.rb
charvp: I was thinking about making this configurable, if someone else wants to do a deploy. I'll do that. For us three days is definitely fine, since we have quite a lot of space left. Transcoded files are also usually a lot smaller than their source material.

config/initializers/delayed_job.rb
charvp: It was only used in the migration, right? If we ever need to run that again it will probably be from the console (where choosing the priority manually is probably for the better). Only if we ever expose it via the API should a queue be added here, IMO.

db/migrate/20200720092916_add_cron_to_delayed_jobs.rb
charvp: Sure, I'll edit. Overlap with usage shouldn't matter though.

charvp:
> Some tests (if possible) would be nice to keep our coverage the same level wink

Tests are rather hard to write for the `audio` action of the TracksController, sadly.

app/jobs/transcode_cache_clean_job.rb
robbevp: 👍

config/initializers/delayed_job.rb
robbevp: Yeah, sure

PR #114, opened 2020-10-01
Add plays model

### LINKED ISSUE #16: Add play count for tracks

robbevp:
fix #16 

Add play model to create a play count. Plays are saved individually, so that we can create more interesting lists and graphs in the long run.
I've not added any other methods than create, since it currently wouldn't make much sense to update, delete, ... these plays.

Some ideas that I haven't included yet:
* don't allow `played_at` to be in the future (though we will have to figure out time zones)
* make sure that a user can't create a play for a different user (

I was doubting whether to send an array of all plays with a track. The other option would be to include a array with `{ :user_id => :play_count }` for each user. At the moment, doing this calculation would be more costly, but there might be a point where the array of individual plays will become too big* to be efficient. Maybe some caching is in order?
(In web, I would start with showing a personal play count and/or a total play count per track. Exposing the individual plays doesn't seem very interesting to me.

*It depends on how much tracks are played, but with a few users that have a very consistent listening pattern, the table could grow pretty quickly.


charvp (changes requested):
> Some ideas that I haven't included yet:
>     * don't allow `played_at` to be in the future (though we will have to figure out time zones)

No need to figure out timezones, we should always use datetime representations that include the timezone.

>     * make sure that a user can't create a play for a different user

This is not an "idea", IMO, but very much a requirement for merging. The fix is easy though, just set the `user_id` to the id of the current user instead of allowing the frontend to set it.

> I was doubting whether to send an array of all plays with a track. The other option would be to include a array with `{ :user_id => :play_count }` for each user. At the moment, doing this calculation would be more costly, but there might be a point where the array of individual plays will become too big* to be efficient. Maybe some caching is in order?

I would think we only want to send the play count of the current user for a track? If/when we start to do some more "social network"-y stuff, we definitely need to think about privacy implications, etc.

robbevp:
> > Some ideas that I haven't included yet:
> > * don't allow `played_at` to be in the future (though we will have to figure out time zones)
> 
> No need to figure out timezones, we should always use datetime representations that include the timezone.
> 

I've added this (and a test for it).

> > ```
> > * make sure that a user can't create a play for a different user
> > ```
>
> This is not an "idea", IMO, but very much a requirement for merging. The fix is easy though, just set the `user_id` to the id of the current user instead of allowing the frontend to set it.
> 
Good idea, implemented

> > I was doubting whether to send an array of all plays with a track. The other option would be to include a array with `{ :user_id => :play_count }` for each user. At the moment, doing this calculation would be more costly, but there might be a point where the array of individual plays will become too big* to be efficient. Maybe some caching is in order?
> 
> I would think we only want to send the play count of the current user for a track? If/when we start to do some more "social network"-y stuff, we definitely need to think about privacy implications, etc.

I'm okay with just sending a play count for that user as a start, and have implemented this.
But don't really see how there would be privacy implications. I think we can assume that the users of one deploy will always be a small group that already know/trust each other. Or do you think they would not want the other users to knwo what they listened to?

robbevp:
Rebased and should be ready for review

app/controllers/tracks_controller.rb
```diff
@@ -11,6 +11,9 @@ class TracksController < ApplicationController
   def index
     authorize Track
     @tracks = apply_scopes(policy_scope(Track))
+              .joins("LEFT JOIN plays ON tracks.id = plays.track_id AND plays.user_id = #{current_user.id}")
```
charvp, line 14: Even though `current_user.id` is not (really) user-controlled, string interpolation into raw SQL is a big no-no.

app/models/play.rb
```diff
@@ -0,0 +1,20 @@
...
+#  played_at :datetime         not null
+#  track_id  :bigint           not null
+#  user_id   :bigint           not null
+#
+class Play < ApplicationRecord
+  belongs_to :track
+  belongs_to :user
+
+  validates :played_at, presence: true
+  validate :play_cannot_be_in_future
```
charvp, line 15: I'm not entirely sure about this validation. It seems logical in theory, but slight clock drifts can make this very annoying for users. I also don't think faking plays to be in the future is something we should be worried about?

app/models/play.rb
```diff
@@ -0,0 +1,20 @@
...
+# Table name: plays
+#
+#  id        :bigint           not null, primary key
+#  played_at :datetime         not null
+#  track_id  :bigint           not null
+#  user_id   :bigint           not null
+#
+class Play < ApplicationRecord
+  belongs_to :track
+  belongs_to :user
```
charvp, line 12: No reverse relation in `user.rb`?

app/policies/play_policy.rb
```diff
@@ -0,0 +1,15 @@
+class PlayPolicy < ApplicationPolicy
+  class Scope < Scope
```
charvp, line 2: I would leave out the scope until it is something we actually use.

app/controllers/plays_controller.rb
```diff
@@ -0,0 +1,18 @@
+class PlaysController < ApplicationController
+  def create
+    authorize Play
+    @play = Play.new(transformed_attributes)
+
+    if @play.save
+      render json: @play, status: :created
+    else
+      render json: @play.errors, status: :unprocessable_entity
```
charvp, line 9: I feel like this line could be tested?

app/serializers/track_serializer.rb
```diff
@@ -32,4 +32,9 @@ def bitrate
   def location_id
     object.audio_file&.location_id
   end
+
+  def play_count
+    # This fallback will only be accessed if the track was just created
```
charvp, line 37: If so, setting the attribute in the track constructor seems cleaner to me.

db/schema.rb
```diff
@@ -193,6 +193,14 @@
     t.index ["path"], name: "index_locations_on_path", unique: true
   end
 
+  create_table "plays", force: :cascade do |t|
```
charvp, line 196: This table will probably be the largest table in a lot of setups. Since we will almost always query this by both user and track, it might be worth it to set an index combining `track_id` and `user_id`.

charvp (changes requested):

app/serializers/track_serializer.rb
robbevp: It would be cleaner, but `play_count` depends on the `current_user`, which we cannot access in the model.

charvp: I don't really get your comment? You don't use the `current_user` here either? I just mean that the fallback value should be set in the constructor, not that the value should be calculated in the constructor.

app/controllers/tracks_controller.rb
robbevp: I've recreated this with Arel

charvp: Can't you write `.joins("LEFT JOIN plays ON tracks.id = plays.track_id AND plays.user_id = ?", current_user.id)`? Most activerecord query methods support that syntax.

charvp: Or maybe a `.where("plays.user_id = ?", current_user.id)` after the join would be cleaner. I feel like that should give the same result?

robbevp: Your suggestions do not work. Postgres throws a weird `RuntimeError (unknown class: Integer)` without a traceback.

Where certainly does not work. A `where`-filter would only return those track that have plays.

app/controllers/plays_controller.rb
robbevp: added

charvp:
I was playing around with the join, and noticed that the tests don't test that only plays of the current user are counted. It seems like an important thing to test though.

robbevp:
> I was playing around with the join, and noticed that the tests don't test that only plays of the current user are counted. It seems like an important thing to test though.

Test added. But currently still fails

robbevp:
I've tried a few more things to get a correct query, by none of these seem to work. [Depending on arel_table and other private api's from rails](https://discuss.rubyonrails.org/t/what-has-happened-to-arel/74383/3) also isn't a very futureproof method.

Two options I can see:
* Deal with the risk (we could always wrap this in `Arel.sql` but I don't think it would make much difference). Since current_user is set when a user is authenticated it would be next to impossible to inject something.
* Don't filter the plays for the current_user. As mentioned before I don't see that much issue with having all plays be public to all users of a certain deploy. If we don't filter these, we bypass the issue altogether

robbevp:
Reworked to avoid the difficult/complex query to get the playcount for the current user.

We now serve plays through PlaysController#index instead of with the tracks.

This has some benefits:
* Less complex query's
* Rendering and serving the tracks won't be slowed down by the plays once a deploy gets a large numbers of plays.
* Handling the plays separately from tracks will make sure we don't slow down handling tracks in the front-end applications.

Since I can see good reasons to keep plays public and private. I've added this to `config/application` so each deploy can decide for themselves (and easily switch afterwards).

robbevp:
@chvp rebased and fixed rubocop issue

test/controllers/plays_controller_test.rb
```diff
@@ -0,0 +1,34 @@
...
+    @track = create(:track)
+    @user = create(:user)
+    sign_in_as @user
+  end
+
+  test 'should get index with only plays for user' do
+    create(:play, track: @track, user: create(:user))
+    get plays_url
+    assert_response :success
+    body = ActiveSupport::JSON.decode response.body
```
chvp, line 14: What is wrong with `JSON.parse`?

chvp (changes requested):
One tiny question, but one bigger remark: plays are lost when a track is deleted currently. That makes sense of course, but shouldn't we transfer plays when merging?

test/controllers/plays_controller_test.rb
robbevp: We also use ActiveSupport::JSON.decode in `test/controllers/tracks_controller_test.rb`, so I've mirrored/used the same structure for these tests

robbevp:
> One tiny question, but one bigger remark: plays are lost when a track is deleted currently. That makes sense of course, but shouldn't we transfer plays when merging?

I agree that we still need to tackle merging. Do you want to do that in this PR or in a separate one?

test/controllers/plays_controller_test.rb
chvp: The `TracksControllerTest` uses both `JSON.parse` and `ActiveSupport::JSON.decode`. The other tests all use `JSON.parse`. If there is no clear reason to use `JSON.parse`, I think I would prefer that (it's shorter).

chvp:
Feels like something that should be done in this PR to me.

robbevp:
> Feels like something that should be done in this PR to me.

Added this + additional test

chvp (approved):

PR #129, opened 2020-10-16
Add test for transcode cache clean job

robbevp:


charvp (approved):

charvp (changes requested):
The ffmpeg call should be mocked out.

charvp (approved):

PR #144, opened 2020-11-29
Fix error in rescan due to albumartist being empty string

robbevp:
This PR fixes an error that could happen during the rescan if the album has an albumartist that is an empty string.

Previously we would try to create an albumartist with an empty string as name
https://github.com/accentor/api/blob/72c5ce46bd6dc5910020bbf1983ce4b54f5dd2bd/app/models/rescan_runner.rb#L99
This fails the validations, but we only notice this in the end when we try to save the track
https://github.com/accentor/api/blob/72c5ce46bd6dc5910020bbf1983ce4b54f5dd2bd/app/models/rescan_runner.rb#L151

We now explicitly check whether the albumartist is present. We already check this for artist, genres, composer, ... or have a fallback value, so this should be the only tag where this error could happen


charvp (changes requested):
Can you add a test case?

robbevp:
> Can you add a test case?

done

charvp (approved):

PR #188, opened 2021-02-21
Check ffmpeg version and allow validation of ContentLengths

### LINKED ISSUE #175: Changing FFmpeg version should invalidate ContentLength cache
- [x] There are no similar issues reported yet

**Describe the bug**
Changes to the FFmpeg internals can change the encoded length of files, breaking clients depending on this length being correct (like the android app).

**To Reproduce**
Find two ffmpeg version that encode to different sizes with the same settings.

**Expected behavior**
The way to fix this is probably to parse out the FFmpeg version on startup. If we add an `ffmpeg_version` field to the ContentLength cache we can delete the entries that don't have a matching version.

Open question: Do we want to queue jobs for ContentLength creation for the AudioFile/CodecConversion combinations that don't have a matching ContentLength? This probably improves user experience, but queuing `|codec_conversion| * |audio_file|` conversion jobs isn't something we want to do on the regular 😬.

robbevp:
# Description
Fix #175 

This PR adds an initializer to get the version of ffmpeg installed and a ffmpeg_version attribute to all `ContentLength`s to we can keep track of the version.

It also adds a task `ffmpeg:queue_version_checks`, to queue a check for all content lengths. This task should be added to deploy scripts or can be manually triggered.
(The initializers don't automatically have access to the models & delayed_job, so it seemed wrong to try and trigger the check there.).

If the ffmpeg_version has changed, the calculation for new `ContentLengths` is added at the end of the queue. There are currently two config options: when the related track was created and the length of the audio_file.
The default are 1 month ago and 5 minutes, but this can be debated. If we want all `ContentLength`s to be recalculated, we can set the length to 0.

# How Has This Been Tested?

- [x] I've added tests relevant to my changes

### DIFF
--- a/app/models/content_length.rb
+++ b/app/models/content_length.rb
@@ -22,6 +22,13 @@ class ContentLength < ApplicationRecord
   def check_ffmpeg_version
     return true if ffmpeg_version == Rails.application.config.FFMPEG_VERSION
 
+    if audio_file.length > Rails.application.config.recalculate_content_length_if_longer_than ||
+       audio_file.track.created_at.after?(Rails.application.config.recalculate_content_length_if_newer_than.call)
+      CodecConversion.find_each do |cc|
+        audio_file.delay(queue: :content_lengths).calc_audio_length(cc)
+      end
+    end
+
     destroy
     false
   end
--- a/app/models/content_length.rb
+++ b/app/models/content_length.rb
@@ -31,8 +31,6 @@ class ContentLength < ApplicationRecord
         audio_file.delay(queue: :content_lengths_backlog).calc_audio_length(cc)
       end
     end
-
-    false
   end
 
   private
--- a/app/models/content_length.rb
+++ b/app/models/content_length.rb
@@ -3,6 +3,7 @@
 # Table name: content_lengths
 #
 #  id                  :bigint           not null, primary key
+#  ffmpeg_version      :string           not null
 #  length              :integer          not null
 #  audio_file_id       :bigint           not null
 #  codec_conversion_id :bigint           not null
@@ -15,4 +16,11 @@ class ContentLength < ApplicationRecord
   validates :audio_file, presence: true, uniqueness: { scope: :codec_conversion }
   validates :codec_conversion, presence: true
   validates :length, presence: true
+
+  before_save :set_ffmpeg_version
+  private
+
+  def set_ffmpeg_version
+    self.ffmpeg_version = Rails.application.config.FFMPEG_VERSION
+  end
 end
--- a/app/models/content_length.rb
+++ b/app/models/content_length.rb
@@ -27,7 +27,7 @@ class ContentLength < ApplicationRecord
     if audio_file.length > Rails.application.config.recalculate_content_length_if_longer_than ||
        audio_file.track.created_at.after?(Rails.application.config.recalculate_content_length_if_newer_than.call)
       CodecConversion.find_each do |cc|
-        audio_file.delay(queue: :content_lengths).calc_audio_length(cc)
+        audio_file.delay(queue: :content_lengths_backlog).calc_audio_length(cc)
       end
     end
 
--- a/app/models/content_length.rb
+++ b/app/models/content_length.rb
@@ -25,11 +25,10 @@ class ContentLength < ApplicationRecord
 
     destroy
 
-    if audio_file.length > Rails.application.config.recalculate_content_length_if_longer_than ||
-       audio_file.track.created_at.after?(Rails.application.config.recalculate_content_length_if_newer_than.call)
-      CodecConversion.find_each do |cc|
-        audio_file.delay(queue: :content_lengths_backlog).calc_audio_length(cc)
-      end
+    return unless Rails.application.config.recalculate_content_length_if.call audio_file
+
+    CodecConversion.find_each do |cc|
+      audio_file.delay(queue: :content_lengths_backlog).calc_audio_length(cc)
     end
   end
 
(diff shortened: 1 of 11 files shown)


robbevp:
@chvp Do you have an idea how long it would take to recreate these `ContentLength`s?
If this takes a long time, we might want add an order when adding these checks to the queues

chvp:
The full recalculation for our library from when I created #175 took about 3-4 days.

robbevp:
Do you have an estimate of what % of track we would get with the current filter (audio > 5min & track created < 1 month ago)? 

Think we best add an order to handle the longest tracks first (since those would cause the user the most wait time). We should probably also distinguish between the backlog of calculations and calculations triggered by a user request - now they would both end up in the same queue

robbevp:
Do you have an estimate of what % of track we would get with the current filter (audio > 5min & track created < 1 month ago)? 

Think we best add an order to handle the longest tracks first (since those would cause the user the most wait time). We should probably also distinguish between the backlog of calculations and calculations triggered by a user request - now they would both end up in the same queue

robbevp:
Seems `find_each` only works with the primary key, so we cannot add them to the queuein order of length.

config/application.rb
```diff
@@ -35,5 +35,7 @@ class Application < Rails::Application
     config.active_job.queue_adapter = :delayed_job
 
     config.transcode_cache_expiry = -> { 1.day.ago }
+    config.recalculate_content_length_if_longer_than = 299
+    config.recalculate_content_length_if_newer_than = -> { 1.month.ago }
```
chvp, line 39: I would just make this a proc that takes an audio file and returns a boolean.

config/initializers/ffmpeg.rb
```diff
@@ -0,0 +1,10 @@
+require 'open3'
+
+if Rails.env.test?
+  # We don't require test setups to have FFMPEG installed, so can jsut manually set a version
```
chvp, line 4:
```suggestion
  # We don't require test setups to have FFMPEG installed, so can just manually set a version
```

app/models/content_length.rb
```diff
@@ -15,4 +16,27 @@ class ContentLength < ApplicationRecord
...
+    destroy
+
+    if audio_file.length > Rails.application.config.recalculate_content_length_if_longer_than ||
+       audio_file.track.created_at.after?(Rails.application.config.recalculate_content_length_if_newer_than.call)
+      CodecConversion.find_each do |cc|
+        audio_file.delay(queue: :content_lengths_backlog).calc_audio_length(cc)
+      end
+    end
+
+    false
```
chvp, line 34: Is this return value used anywhere?

app/models/content_length.rb
```diff
@@ -15,4 +16,27 @@ class ContentLength < ApplicationRecord
   validates :audio_file, presence: true, uniqueness: { scope: :codec_conversion }
   validates :codec_conversion, presence: true
   validates :length, presence: true
```
chvp, line 18: `validates` for `ffmpeg_version` should be added.

lib/tasks/ffmpeg.rake
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
```
chvp, line 2: I would leave this out. This will be the default in ruby 3 and there are no strings in this file.

lib/tasks/ffmpeg.rake
```diff
@@ -0,0 +1,9 @@
...
+namespace :ffmpeg do
+  task queue_version_checks: :environment do
+    ContentLength.find_each do |cl|
+      cl.delay(queue: :content_lengths).check_ffmpeg_version
```
chvp, line 6: Doesn't using the backlog queue make more sense?

test/models/content_length_test.rb
```diff
@@ -11,4 +12,74 @@
...
+  end
+
+  test 'should not create new ContentLength if audio is longer than config and track is older than config' do
+    @track = create(:track, audio_file: @audio_file)
+
+    # rubocop:disable Rails/SkipsModelValidations
+    # We want to avoid our own before_save callbacks to manually set a wrong version and audio_file length
+    @audio_file.update_column(:length, 1)
+    @audio_file.track.update_column(:created_at, 1.year.ago)
+    @content_length.update_column(:ffmpeg_version, Faker::App.semantic_version)
```
chvp, line 41: This generated version can be the same as `0.0.1`. This can result in tests that fail at random.

chvp (changes requested):

chvp:
About a third of the audio files in our library are caught by the current filters (the vast majority of which through the longer than 5 minutes rule).

test/models/content_length_test.rb
robbevp: Changed the version in the initializer to '0.0.0' - faker always has a patch between [1 to 9](https://github.com/faker-ruby/faker/blob/master/lib/faker/default/app.rb)

lib/tasks/ffmpeg.rake
robbevp: We need to verify a whether the content_lengths can be used quicker than we need the content_lengths themselves, so I would give these two tasks a different priority

lib/tasks/ffmpeg.rake
robbevp: It wasn't included in ruby 3 [after all](https://bugs.ruby-lang.org/issues/11473#note-53) but fair point nonetheless

robbevp:
> About a third of the audio files in our library are caught by the current filters (the vast majority of which through the longer than 5 minutes rule).

That seems reasonable to me. We can always tweak the limits for recalculation if necessary

robbevp:
superseded by #203

PR #199, opened 2021-04-10
Add gitattributes

robbevp:
## Description

Add .gitattributes to help github with language detection.
This is based on the default that ships with new rails apps: https://github.com/rails/rails/blob/291a3d2ef29a3842d1156ada7526f4ee60dd2b59/railties/lib/rails/generators/rails/app/templates/gitattributes.tt

I've excluded `yarn.lock` since this doesn't apply here.
I've excluded `vendor/*` since these aren't included in git.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# See https://git-scm.com/docs/gitattributes for more about git attribute files.
+
+# Mark the database schema as having been generated.
+db/schema.rb linguist-generated


chvp:
If I understand things correctly marking files as generated mostly means their diffs won't be shown by default in pull requests. While it is true that the schema definition file is generated, it is a file where I look at the changes if a pull requests changes that file. So I don't think marking it as something where the diff should not be shown makes sense.

robbevp:
Fair point. It would indeed be hidden by default, as the yarn.lock file is now: https://github.com/accentor/web/pull/415/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de
We could also mark it as [documentation](https://github.com/github/linguist/blob/master/docs/overrides.md#summary) or just close this if that doesn't seem like a good fit either

chvp:
I don't think marking it as documentation makes a lot of sense conceptually.

PR #203, opened 2021-04-25
Check ffmpeg version and recalculate ContentLengths

### LINKED ISSUE #175: Changing FFmpeg version should invalidate ContentLength cache

robbevp:
# Description
Fix #175
Supersedes #188

A different take on the issue of ffmpeg versions - the responsibility to trigger check is now clearly up to the person who deploys the application.
The version check is handled by a rake task `ffmpeg:check_version`. This checks whether the stored version (either a file supplied as an envitionment variables or `log/ffmpeg_version.txt`) still matches the current version, triggers a method if not. If the file cannot be found or read, we fallback to version 0.0.0.

There is also a helper provided: `ffmpeg:init`. This sets the current ffmpeg version to the supplied environment variable or to `log/ffmpeg_version.txt`

If the versions do not match we destroy all content lengths and recalculate them for each AudioFile that matches the condition in `config/application.rb`

# How Has This Been Tested?

- [x] I've added tests relevant to my changes


config/environments/test.rb
```diff
@@ -47,9 +47,9 @@
   config.active_job.queue_adapter = :delayed_job
   Delayed::Worker.delay_jobs = false
 
-  # For tests we want to be the transcode cache expiry setting to remain
-  # the same, regardless of the configuration in application.rb
+  # For tests we want to be the these settings to remain the same, regardless of the configuration in application.rb
```
chvp, line 50:
```suggestion
  # For tests we want these settings to remain the same, regardless of the configuration in application.rb
```

chvp (changes requested):

chvp (approved):

PR #206, opened 2021-05-02
Fix annotate route generation

robbevp:
## Description
Since rails 6.1 the rake commen to generate routes if no longer available, but the annotate gem still depends on this task.
This PR applies a small workaround, that we can keep until annotate is updated.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,8 +52,7 @@ group :development, :test do
 end
 
 group :development do
-  gem 'annotate', '~> 3.1'
-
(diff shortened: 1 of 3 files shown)


chvp (changes requested):
A comment should probably be added to the Gemfile, to remind ourselves to remove any workarounds when updating `annotate`.

robbevp:
> A comment should probably be added to the Gemfile, to remind ourselves to remove any workarounds when updating `annotate`.

Good idea. Added

chvp (approved):

PR #226, opened 2021-07-11
Rename merge params to `source_id`

robbevp:
Makes the merge param consistent accross all controllers.
Necessary for the` api-client-js` to work

- [x] I've ~~added~~ updated tests relevant to my changes.

### DIFF
--- a/app/controllers/artists_controller.rb
+++ b/app/controllers/artists_controller.rb
@@ -48,7 +48,7 @@ class ArtistsController < ApplicationController
   end
 
   def merge
-    render json: @artist.errors, status: :unprocessable_entity unless @artist.merge(Artist.find(params[:old_id]))
+    render json: @artist.errors, status: :unprocessable_entity unless @artist.merge(Artist.find(params[:source_id]))
   end
 
   private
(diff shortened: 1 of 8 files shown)


chvp (commented):
I'm not sure about the `old` naming. There is no clear time order when merging. (I'm not actually sure if it's the older item that will be removed when merging, but my intuition says no.)

robbevp:
The second param (so `old_id` or `other_MODEL_id`) is the item that gets destroyed.
I would like to have a param that makes it clear which item gets destroyed, but agree that `old_id` isn't fully clear either. Any suggestions?

chvp:
`removed_id` maybe? It's not perfect, since the item isn't removed yet at the time of calling, but `to_remove_id` feels a little clunky.

chvp:
Also, can you open a PR on web that incorporates these changes? I'd like to merge them at the same time.

robbevp:
Maybe `source_id`? (where the first id would be the `target`)

chvp:
`source_id` sounds good.

chvp (approved):

PR #244, opened 2021-08-21
Make gitignore for `/vendor/*` explicit

robbevp:
I removed this from my global gitignore for another project. Since we don't want these files to show up in git, it seems best to make this explicit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 # Ignore bundler config.
 /.bundle
 
+# Ignore files from vendors
+/vendor/*
+
 # Ignore all logfiles and tempfiles.
 /log/*
 /tmp/*


chvp (commented):
Why would we not want these files (which files?) to show up in git? If we ever vendor a dependency, we do want to track that in git.

robbevp:
The vendor folder (at least on my machine) contains all gems, extentions, ... installed by bundler. If we would ever want to track a dependency, we can always update gitignore to include that specific directory

chvp:
Can you not just add the subdirectory you use for bundler installed stuff to your global gitignore. Mine contains `/vendor/bundle`, allowing me to do the same thing without having my setup impact the gitignore files of the projects I work on.

robbevp:
sure, that works too

PR #245, opened 2021-08-21
Allow plays to be scoped by album

robbevp:
Add a simple scope to get plays by the album they are related too


chvp (changes requested):
Please add some tests.

chvp (approved):

PR #249, opened 2021-09-04
Add rake task to start rescan

chvp:
No tests, since that's not really a thing for rake tasks AFAIK. I did test locally though (started a worker and then executed the task, then verified that the worker started the job).

### DIFF
--- a/lib/tasks/rescan.rake
+++ b/lib/tasks/rescan.rake
@@ -1,6 +1,5 @@
 namespace :rescan do
   task start: :environment do
-    rescan = RescanRunner.first || RescanRunner.create
-    rescan.delay(queue: :rescans).run
+    RescanRunner.instance.start_delayed
   end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/rescan.rake
```diff
@@ -0,0 +1,6 @@
+namespace :rescan do
+  task start: :environment do
+    rescan = RescanRunner.first || RescanRunner.create
```
robbevp, line 3: I feel like more of this look could live inside the model? Now we have the same code in the controller and here.

In another project we have a public `Runner.schedule` and keep run private, so we're sure it can never be called outside of a delay queue. (Though there we have multiple runners and scheduling is a bit more involved)

lib/tasks/rescan.rake
```diff
@@ -0,0 +1,5 @@
+namespace :rescan do
+  task start: :environment do
+    RescanRunner.instance.start_delayed
```
robbevp, line 3: Should probably be `.schedule` instead of `.start_delayed`?

robbevp (changes requested):

robbevp (approved):

PR #250, opened 2021-09-05
Runner per location

### LINKED ISSUE #150: Handle rescan per location
**Is your feature request related to a problem? Please describe.**
The larger a deploy's collection is getting, the longer it takes to scan the whole collection.
Usually an admin/moderator will know which files are new or deleted. While it isn't bad to check all other files for something new or deleted, it will take more and more time.

This would be especially frustrating to a deploy with multiple locations: A user knows that files where deleted/added in one location, but has to wait until all locations are scanned. (Imagine that our current setup has a location per user instead of one folder with eveything)

**Describe the solution you'd like**
I would split up the rescan per location: where every location has one rescan object. 

This way a rescan can be triggered per location, so the user can scan just that location where files where added/removed. Or a scan for each location can be added to a queue.

Of course, this would cause a heavy rework of the rescan model and breaking changes to the API routes.

robbevp:
Fixes #150 

This PR changes the behaviour of the `RescanRunner` to work per location. This will offer two benefits that can both speed up scanning:
* Parallelization (if there are multiple locations and multiple workers, that is)
* More atomic rescans (since an admin/mod usually knows where files were changed)

This causes a breaking change in routes, so we mention this in the release notes and refer to a version of web that is compatible with these changes.

### Things to discuss:
1. We currently lose the warning when no locations exist. We might want to add this to the controller?
2. While you can't schedule a run while that runner is active, but you could while the job is waiting to be picked up. We might want to change the `running?` boolean to an enum with inactive, scheduled and running, so that we're sure that a runner isn't scheduled twice?


config/routes.rb
```diff
@@ -144,8 +147,8 @@
       end
     end
     resources :users
-
-    get '/rescan' => 'rescan#show'
-    post '/rescan' => 'rescan#start'
+    resources :rescans, only: %i[index show]
+    post 'rescans/:id', to: 'rescans#start'
+    post 'rescans', to: 'rescans#start_all'
```
chvp, line 152: I feel like it should be possible to express this a bit more cleanly in the DSL.

chvp (commented):
Can you make a matching PR in api-client-js first? I'd like to see this new API in action.

config/routes.rb
robbevp: There might be, but I couldn't find it (at least if we want these paths). 
If we want to add these routes as extra actions on the resource, we'll always end up with something like this:

```ruby
resources :rescans, only: %i[index show] do
    post 'start', on: :member
    post 'start_all', on: :collection 
end

# start_rescan POST   /api/rescans/:id/start(.:format)                                                                  rescans#start
# start_all_rescans POST   /api/rescans/start_all(.:format)                                                                  rescans#start_all
```

test/models/location_test.rb
```diff
@@ -9,4 +9,9 @@
 require 'test_helper'
 
 class LocationTest < ActiveSupport::TestCase
+  test 'should create a rescan runner' do
+    assert_difference('RescanRunner.count') do
```
chvp, line 13:
```suggestion
    assert_difference('RescanRunner.count', 1) do
```
I know that this is the default, but I prefer to be explicit about it, personally (feel free to disagree).

chvp (approved):
Will need to be rebased after #266 and #250.

chvp (changes requested):
Actually, when testing locally, I noticed that the seeds need to be updated.

test/models/location_test.rb
robbevp: I don't care that much so i've updated this. Can you check if there is a rubocop cop for this and enable it? It's more important to me that this is consistent

robbevp:
@chvp rebased and seeds updated

chvp (approved):

chvp:
> 1. We currently lose the warning when no locations exist. We might want to add this to the controller?

Seems like something that could be handled client-side.

> 2. While you can't schedule a run while that runner is active, but you could while the job is waiting to be picked up. We might want to change the `running?` boolean to an enum with inactive, scheduled and running, so that we're sure that a runner isn't scheduled twice?

Sounds like a good idea, but also sounds like something for a subsequent PR.

PR #254, opened 2021-09-19
Verify new Location is not parent or subdir of existing path

robbevp:
Verifies whether a new location is a parent or subdirectory of an existing location.

I think it makes sense to add this validation, since adding two paths that contain the same files will cause tracks to show up twice.

The validation is added in the simplest way possible, by checking whether the DB already has path that either starts with the new path or where new path starts the same as an existing string. This obviously has two drawbacks:
* If a relative path is added, this won't work (though I'm not sure what if relative paths work for the RescanRunner)
* If the server admin uses symlinks, we might miss some things.


chvp (changes requested):
I'm not sure the casing thing makes sense. Paths on sane OS's/filesystems are case-sensitive, so `/var` and `/Var` are different things.

chvp (approved):

chvp:
I was a bit too eager with merging this. The following test does not succeed:
```ruby

  test 'should be able to add similar siblings' do
    sibling = build(:location, path: '/var/parent2')
    assert sibling.valid?
  end
```

PR #273, opened 2021-11-27
Add hidden attribute to TrackArtist

robbevp:
Allows mod+ to "hide" a TrackArtist. This way, detailed credits can be added to a song, while the main overviews will stay minimal (if that is what is wanted for a track).

No additional tests added - since I couldn't think of a relevant test here.

### DIFF
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -140,7 +140,7 @@ class TracksController < ApplicationController
 
     if attributes[:track_artists].present?
       attributes[:track_artists] = attributes[:track_artists].map do |ta|
-        TrackArtist.new(artist_id: ta[:artist_id], name: ta[:name], role: ta[:role], order: ta[:order] || 0, hidden: ta[:hidden])
+        TrackArtist.new(artist_id: ta[:artist_id], name: ta[:name], role: ta[:role], order: ta[:order] || 0, hidden: ta[:hidden] || false)
       end
     end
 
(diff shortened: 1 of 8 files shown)


robbevp:
@chvp Do you see what is going with the tests?
I just get:
Screenshot 2021-11-27 at 16 10 18
 
The test suite worked locally

db/migrate/20211127114919_add_hidden_to_track_artists.rb
```diff
@@ -0,0 +1,5 @@
+class AddHiddenToTrackArtists < ActiveRecord::Migration[6.1]
+  def change
+    add_column :track_artists, :hidden, :boolean, default: false
```
chvp, line 3: Also `null: false`?

app/controllers/tracks_controller.rb
```diff
@@ -140,7 +140,7 @@ def transformed_attributes
 
     if attributes[:track_artists].present?
       attributes[:track_artists] = attributes[:track_artists].map do |ta|
-        TrackArtist.new(artist_id: ta[:artist_id], name: ta[:name], role: ta[:role], order: ta[:order] || 0)
+        TrackArtist.new(artist_id: ta[:artist_id], name: ta[:name], role: ta[:role], order: ta[:order] || 0, hidden: ta[:hidden])
```
chvp, line 143: Does this require clients to pass the `hidden` parameter? What happens if they don't? (Maybe add some tests to confirm the expected behavior?)

chvp (changes requested):

app/controllers/tracks_controller.rb
robbevp: It would have been `false`, since that was set as default. Since I also added `null: false` to the column, I had to add a fallback value

chvp (approved):

PR #274, opened 2021-12-04
Fix migration for rescan runner per location

chvp:
The migration didn't work, because `destroy_all` went looking for the relevant location (since there is a `belongs_to` in the model).


robbevp (approved):
👍

Not sure if I wrote the migration before the changes to the model or just migrated back-and-forth a few times and missed things that way. Either way, this seems a good fix

PR #282, opened 2022-01-15
Try to match a list of genres in rescan

robbevp:
This PR improves the way the rescan runner handles the genre tag and allows creating multiple genres from one tag.

Since ID3 tag and vorbis tags only allow one genre field, there are a lot of files that have a list of genres inside that one tag. Up until now we created a genre with the literal string inside the tag, often resulting in genres like `Folk/Rock` or `Rock, Folk, Folk-Rock` - which then need a manual conversion to multiple genres.

With this PR, this logic is changed so that we pick up a list of genres and create a genre for each item in the list.

The logic is now as follows:
If the genre is already in the database (even if it is a list), we match this genre. (So that a user can use the separator inside a genre name if they want).
Otherwise, we split the string by `,` or `/` and try to match each part (normalized and stripped of whitespace to allow `Genre 1, Genre 2`) or create new genre for each part

There might be some other separators used or relevant, but these two seemed to be the two most common in our library. A combination of both is also possible (however unlikely/messy that sounds)

The new file in the test contains the genre tag `genre 1, genre 2 / genre 3`


app/models/rescan_runner.rb
```diff
@@ -184,6 +183,19 @@ def find_cover(path)
...
   end
 
+  def match_genres(tag)
+    return [] if tag.blank?
+
+    # If the exact genre can be found, we always use that
+    genre = Genre.find_by(normalized_name: Genre.normalize(tag))
+    return [genre] if genre.present?
+
+    # Split tag by comma or dash and match or create genre for each part
```
chvp, line 193:
```suggestion
    # Split tag by comma or slash and match or create genre for each part
```

chvp (approved):

PR #283, opened 2022-01-22
Update to ruby 3.1

chvp:


robbevp (approved):
Just to be sure: did you do some manual tests as well? Rails 7 had some issues with ruby 3.1 (which were fixed in Rails 7.0.1). Not sure if the same was the case for Rails 6.1

PR #313, opened 2022-03-26
Calc play count per user in SQL view

robbevp:
In the web app, we currently fetch all play data quite a lot:
* With each global reload
* Every time the user goes to an album page

While we do very little with it. On most pages we only display the play count for each track. The stats pages are the only ones where we actually need the plays themselves, to do more finegrained calculations.

In the initial implementation of the plays, we didn't include the play count with the tracks, since we didn't find a good way to include this data only for the current user.

In this PR:
* [x] I added [scenic](https://github.com/scenic-views/scenic) to easily work with PostgreSQL's (materialized) views - it provides an easy way to manage the views in migrations and manage different versions of a view and then consume those views as if they were regular models/tables
* [x] Created a view to calculate a play_count per user/track
* [x] Used that view to include the play_count for the current user with the tracks

### Load/responses times
I currently created a view, which runs each time it is accessed. We'll have to see if this doesn't cause too much load or longer response times for the tracks.
On my dev machine (with ±150 tracks and ±40000 plays) I could load the whole view in ±30ms. There also was no notable difference in the response times of the `TracksController` (Note that these views will only calculate the relevant data. So `User.first.play_counts` should generally be faster than `PlayCount.all`. In the case of the TracksController, this should mean that only the play counts for the 500 tracks that will be returned are calculated on each request.)
If this turns out to cause too much load/delay, we could switch to a materialized view, but then we also have to figure out how and when to refresh the view.

### Including play_count in TrackSerializer
When I started working on this, I did not think that it would be this complex to include the play_counts with in the TracksSerializer.
I assumed that
```ruby
Track.includes(:play_counts).where(play_counts: { user: current_user }).first.play_counts.length
# => 1
```
but it seemes that where clause is left out when we access the play_counts again, so we actually get the number of users.

That means we currently have to make the `TrackSerializer` aware of the current_user:
https://github.com/accentor/api/blob/7627202bac34a8d8f284bf80e2f96466a27a8d50/app/controllers/tracks_controller.rb#L2
And that we need a method that can efficiently load the play_count for a user, regardless of whether we have eager loaded the play_counts or not
https://github.com/accentor/api/blob/7627202bac34a8d8f284bf80e2f96466a27a8d50/app/models/track.rb#L39-L42

Maybe we shouldn't be returning these counts together with the tracks at all, but just through a separate endpoint? But then we start losing some of the advantages of this PR (namely, less API requests)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,9 @@ gem 'bootsnap', '~> 1.11.1', require: false
 # Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin AJAX possible
 gem 'rack-cors', '~> 1.1'
 
+# Use Scenic to easily create (materialized) views in postgresql
+gem 'scenic', '~> 1.6'
+
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -218,6 +218,9 @@ GEM
     ruby-progressbar (1.11.0)
     ruby-vips (2.1.4)
       ffi (~> 1.12)
+    scenic (1.6.0)
+      activerecord (>= 4.0.0)
+      railties (>= 4.0.0)
     simplecov (0.21.2)
       docile (~> 1.1)
       simplecov-html (~> 0.11)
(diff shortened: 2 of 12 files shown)


robbevp:
Just realized that if we would create a separate endpoint for play counts, we would probably be better off to also allow query params to filter by period. This could then also mitigate a lot of the performance problems with the stats page in the web app.

robbevp:
Superseded by #340

PR #314, opened 2022-03-27
Add playlists and playlist items

robbevp:
This PR adds a basic implementation of Playlists and exposes this through the API.

The concept of a Playlist is currently the following:
* A playlist is a collection of either albums, artists or tracks. All items within one playlist should be of the same type.
  This allows us to create playlist like 'My favorite songs', 'My top 10 albums of 2022', 'All artists playing at festival X'
* The items in a playlist are ordered (though a user can choose to not use this).
* A playlist is either personal or shared. A personal playlist can only be updated or destroyed by the user it belongs to, a shared playlist can be updated all the users.
  My primary reasoning for this distinction is for subjective and objective lists. 'Robbe's top 10 ambient albums' would be a personal list. 'All artists playing at Rewire 2022' is an objective list that everyone can add to. Another option for a shared list could be 'Music for our road trip'
* A playlist has a name and a description. This allows the users to keep the name short 'Rewire 2022' and provide extra info in the description 'All artists playing at Rewire 2022'
  
### Private playlists?  
~~I was doubting to allow a user to make their personal playlists private. Currently all users can see all playlists, where a private playlist would only be visible to the user the list belongs to. I didn't think I'd ever mark a playlist private, but if there is a need for it, I wouldn't mind it either.~~

Private playlists are now added

### `PlaylistItem` needed?
We currently store the playlist items in a separate model/table. Mostly so we can use Rails' builtins for associations, and store order as an explicit field.

However, at the moment our only interaction with these items is through the `item_ids` getter/setter, so we don't necessarily need a separate model / table, but also might store `item_ids` in a json/jsonb type column on the playlists table. This would also allow us to remove the checks on the playlist's type and the item's item_type

### DIFF
--- a/app/models/artist.rb
+++ b/app/models/artist.rb
@@ -20,6 +20,8 @@ class Artist < ApplicationRecord
   has_many :tracks, through: :track_artists, source: :track
   has_many :album_artists, dependent: :destroy
   has_many :albums, through: :album_artists, source: :album
+  has_many :playlist_items, as: :item, dependent: :destroy
+  has_many :playlists, through: :playlist_items, source: :playlist
 
   validates :name, presence: true
 
--- a/app/models/artist.rb
+++ b/app/models/artist.rb
@@ -44,6 +44,11 @@ class Artist < ApplicationRecord
       ta.update(artist_id: id)
     end
 
+    # rubocop:disable Rails/SkipsModelValidations
+    # Since we only update the item_id, there aren't any validations that could fail
+    other.playlist_items.update_all(item_id: id)
+    # rubocop:enable Rails/SkipsModelValidations
+
     # we have to reload to make sure the track_artists and album_artists relation isn't cached anymore
     other.reload.destroy
   end
--- a/app/models/artist.rb
+++ b/app/models/artist.rb
@@ -44,10 +44,9 @@ class Artist < ApplicationRecord
       ta.update(artist_id: id)
     end
 
-    # rubocop:disable Rails/SkipsModelValidations
-    # Since we only update the item_id, there aren't any validations that could fail
-    other.playlist_items.update_all(item_id: id)
-    # rubocop:enable Rails/SkipsModelValidations
+    other.playlist_items.find_each do |item|
+      item.update(item_id: id) unless playlist_items.where(playlist_id: item.playlist_id).any?
+    end
 
     # we have to reload to make sure the track_artists and album_artists relation isn't cached anymore
     other.reload.destroy
--- a/app/models/playlist.rb
+++ b/app/models/playlist.rb
@@ -18,11 +18,22 @@ class Playlist < ApplicationRecord
 
   validates :name, presence: true
 
+  before_save :normalize_item_order
+
   def shared?
     user_id.nil?
   end
 
   def personal?
     user_id.present?
+  end
+
+  private
+
+  def normalize_item_order
+    items.sort { |i1, i2| i1.order <=> i2.order }.map.with_index(1) do |item, index|
+      item.order = index
+      item.save unless item.new_record?
+    end
   end
 end
--- a/app/models/playlist.rb
+++ b/app/models/playlist.rb
@@ -3,32 +3,24 @@
 # Table name: playlists
 #
 #  id            :bigint           not null, primary key
+#  access        :integer          default("shared")
 #  description   :string
 #  name          :string           not null
 #  playlist_type :integer          not null
-#  private       :boolean          default(FALSE)
 #  created_at    :datetime         not null
 #  updated_at    :datetime         not null
-#  user_id       :bigint
+#  user_id       :bigint           not null
 #
 class Playlist < ApplicationRecord
-  belongs_to :user, optional: true
+  belongs_to :user
   has_many :items, class_name: 'PlaylistItem', dependent: :destroy
 
+  enum access: { shared: 0, personal: 1, secret: 2 }
   enum playlist_type: { album: 1, artist: 2, track: 3 }
 
   validates :name, presence: true
-  validates :user_id, presence: true, if: :private?
 
   before_save :normalize_item_order
-
-  def shared?
-    user_id.nil?
-  end
-
-  def personal?
-    user_id.present?
-  end
 
   def item_ids
     items.order(:order).pluck(:item_id)
--- a/app/models/playlist.rb
+++ b/app/models/playlist.rb
@@ -18,7 +18,7 @@ class Playlist < ApplicationRecord
   enum access: { shared: 0, personal: 1, secret: 2 }
   enum playlist_type: { album: 0, artist: 1, track: 2 }
 
-  scope :with_item_ids, -> { left_joins(:items).select('playlists.*', 'array_agg(playlist_items.item_id ORDER BY playlist_items.order ASC) as item_ids').group(:id) }
+  scope :with_item_ids, -> { left_joins(:items).select('playlists.*', 'array_remove(array_agg(playlist_items.item_id ORDER BY playlist_items.order ASC), NULL) as item_ids').group(:id) }
 
   validates :name, presence: true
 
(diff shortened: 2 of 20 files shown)


robbevp:
I'm keeping this a draft, since the controller implementation might still change a bit (especially around creating/updating lists), while working on a matching PR in the web release. However, I do think that this is ready to be looked at

app/policies/playlist_policy.rb
```diff
@@ -0,0 +1,31 @@
...
+      scope.where(private: false).or(scope.where(private: true, user_id: user.id))
+    end
+  end
+
+  def index?
+    user.present?
+  end
+
+  def show?
+    user.present?
```
chvp, line 13: Needs to also take into account the `private` flag, no?

app/models/playlist.rb
```diff
@@ -0,0 +1,50 @@
...
+#  description   :string
+#  name          :string           not null
+#  playlist_type :integer          not null
+#  private       :boolean          default(FALSE)
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#  user_id       :bigint
+#
+class Playlist < ApplicationRecord
+  belongs_to :user, optional: true
```
chvp, line 15: Always storing the creator and keeping the three levels of access (private/personal/public) in an enum makes more sense to me, and it feels like it would be harder to mess up access with that system.

app/serializers/playlist_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+#  description   :string
+#  name          :string           not null
+#  playlist_type :integer          not null
+#  private       :boolean          default(FALSE)
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#  user_id       :bigint
+#
+class PlaylistSerializer < ActiveModel::Serializer
+  attributes :id, :name, :description, :user_id, :playlist_type, :created_at, :updated_at, :item_ids, :private
```
chvp, line 15: Have you tested what happens with a huge amount of `item_ids`?

chvp (changes requested):

app/models/playlist.rb
robbevp: I've implemented this, i agree that it makes things a lot simpler. I went for `secret` rather than `private` since an enum can't define the method private (I collides with a method from ActiveRecord).

One extra point here: The user_id now only gets set add creation. This means that if user X makes playlist and then user Y makes it personal or secret, it still belongs to user X. The other option would have been to always set the user_id to the last user who edited the playlist.
I don't think this is a real world problem (since shared playlist already require the trust that another user would not just delete all the items), but wanted to point it out as an edge case

app/serializers/playlist_serializer.rb
robbevp: I just did a few test. There was some N+1 querying going on, but this is now solved.

When we have multiple playlists with a total of 60000 items between them, the request takes between 800ms and 1000ms. This is certainly something to watch out for, if a request would need to return the full 500 lists, all with a >1000 items in each one, this request would probably be painfully slow.

Maybe we need to use a lower number of playlists per page?

robbevp: I added an extra commit, which improves this situation A LOT. The same request with 60000 playlist_items now takes 170 - 300ms and should be able to scale without much issues. It does require more (postgres-only) SQL to be present in the model, but I'd say that is worth it here?

app/models/playlist.rb
chvp: Is access an attribute we want to be able to change after creation? (I don't immediately see a use-case.)

robbevp: I'd say so, mostly because someone might change their mind on something being personal/secret (or just create a playlist a select the wrong access by accident)

app/models/artist.rb
```diff
@@ -42,6 +44,11 @@ def merge(other)
       ta.update(artist_id: id)
     end
 
+    # rubocop:disable Rails/SkipsModelValidations
+    # Since we only update the item_id, there aren't any validations that could fail
+    other.playlist_items.update_all(item_id: id)
+    # rubocop:enable Rails/SkipsModelValidations
+
```
chvp, line 51: Doesn't this mean there could be duplicates in a playlist? I guess that's not a failed validation, but is this something we want in this case?

app/models/playlist.rb
```diff
@@ -0,0 +1,48 @@
...
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#  user_id       :bigint           not null
+#
+class Playlist < ApplicationRecord
+  belongs_to :user
+  has_many :items, class_name: 'PlaylistItem', dependent: :destroy
+
+  enum access: { shared: 0, personal: 1, secret: 2 }
+  enum playlist_type: { album: 1, artist: 2, track: 3 }
```
chvp, line 19: Why not start at 0?

chvp (approved):

chvp:
(Pushed a rebase since this branch didn't include the flake yet.)

app/models/playlist.rb
robbevp: Can't remember - I'll change this

app/models/artist.rb
robbevp: we validate the uniqueness of an item_id within a playlist (both with validations and with an index) - but I'm not sure what would happen here. Probably an error? 

I'll check + add a test to see that the merging will always succeed (and do the same for tracks)

flake.nix
```diff
@@ -58,7 +58,7 @@
             env = [
```
robbevp, line 58: @chvp I think these changes accidentally slipped into this PR when rebasing?

chvp: Right, I was going to make this a separate PR and then I forgot.

robbevp:
Updated the merging strategy of tracks and artists to make sure this works when they are in the same playlist. @chvp Maybe take a short look at the changes

chvp (approved):

robbevp:
Sorry - one more change, since the current result of `with_item_ids` was causing errors in the web interface

app/models/playlist.rb
```diff
@@ -18,15 +18,15 @@ class Playlist < ApplicationRecord
   enum access: { shared: 0, personal: 1, secret: 2 }
   enum playlist_type: { album: 0, artist: 1, track: 2 }
 
-  scope :with_item_ids, -> { left_joins(:items).select('playlists.*', 'array_agg(playlist_items.item_id ORDER BY playlist_items.order ASC) as item_ids').group(:id) }
+  scope :with_item_ids, -> { left_joins(:items).select('playlists.*', 'array_remove(array_agg(playlist_items.item_id ORDER BY playlist_items.order ASC), NULL) as item_ids').group(:id) }
```
chvp, line 21: Can you add a comment in the code on why the `array_remove(..., NULL)` is necessary? It isn't immediately obvious to me now, so it definitely won't be in the future.

app/models/playlist.rb
```diff
@@ -18,15 +18,15 @@ class Playlist < ApplicationRecord
...
 
   validates :name, presence: true
 
   before_save :normalize_item_order
 
   def item_ids
     # If with_item_ids was used, we use the output from the SQL query
-    return read_attribute(:item_ids) if has_attribute?(:item_ids)
+    return read_attribute(:item_ids) || [] if has_attribute?(:item_ids)
```
chvp, line 29: Is there really no way to make the SQL query return an empty array?

chvp (changes requested):

app/models/playlist.rb
robbevp: We could do this with the following, but lose some readability:
```sql
COALESCE(array_remove(array_agg(playlist_items.item_id ORDER BY playlist_items.order ASC), NULL), [])
```

chvp: I think I prefer that solution. The SQL is already complicated enough that it needs some comments IMO, so adding some slight extra complexity there seems preferable to me to adding it here.

robbevp:
@chvp my last commit only contains some reformatting - both before and after should pass our rubocop config, but i felt that a multiline lamdba made more sense here. Feel free to drop this commit if you disagree

chvp (approved):

PR #317, opened 2022-04-02
Nilify empty strings

### LINKED ISSUE #316: Nullify edition description if empty string
We should nullify the edition description (or all strings?) if they are just empty string.

> Would it make more sense to fix this serverside? Feels like empty strings should just not be allowed into the database.
> _Originally posted by @chvp in https://github.com/accentor/web/pull/748#pullrequestreview-929541027_

robbevp:
This PR adds a method to define which values should be nullified if they are blank.

We want to have control over which columns should be normalized, since some columns (`AlbumArtist.separator`) can explicitly be an empty string but not nil.
I've currently added the Album's edition_description and all instances of the review_comment.

Fixes #316 

## Questions
This PR currently doesn't include way to fix this for all current places where the empty string is present. Should we leave this up to the user to fix (in the interface or console), or add a migration to go over all albums, artists and tracks.

I've currently included this method in `ApplicationRecord`, but could also see this being a concern that should be explicitly included in each model

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -16,7 +16,7 @@ class ApplicationRecord < ActiveRecord::Base
       set_callback :validation, :before, method_name, if: col_changed
 
       define_method method_name do
-        self[col].present? || self[col] = nil
+        self[col] = nil if self[col].blank?
       end
     end
   end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -7,4 +7,17 @@ class ApplicationRecord < ActiveRecord::Base
   def self.normalize(str)
     str.unicode_normalize(:nfkd).gsub(/[\u0300-\u036f]/, '').downcase
   end
+
+  def self.normalize_blank_values(*cols)
+    cols.each do |col|
+      col_changed = :"#{col}_changed?"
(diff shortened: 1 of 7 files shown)


app/models/application_record.rb
```diff
@@ -7,4 +7,17 @@ class ApplicationRecord < ActiveRecord::Base
...
+
+  def self.nilify_blank_values(*cols)
+    cols.each do |col|
+      col_changed = :"#{col}_changed?"
+      method_name = :"nilify_blank_#{col}_value"
+
+      set_callback :validation, :before, method_name, if: col_changed
+
+      define_method method_name do
+        self[col].present? || self[col] = nil
```
chvp, line 19: I'm not a fan of using boolean logic for control flow like this. `self[col] = nil if self[col].blank?` is a lot clearer, IMO.

chvp (changes requested):

app/models/application_record.rb
robbevp: agreed. I had a very similar method in a different project, but had to read this part twice to remember how it worked 😅

chvp (approved):

PR #324, opened 2022-04-18
Add nix flake with devshell and package

robbevp:
re Accentor/flake#5

This PR:
* Adds a flake.nix that exposes the api package and a devshell
* Adds a github action to automatically keep gemset.nix in sync (after dependabot or if people don't use nix).

Might be good if this can be tried on different machines and OS's (I only have aarch64-darwin).

### Details
The devshell can be used with `nix develop` or `use flake` inside an `.envrc`-file.

It will print out a message with some of the additional commands:
```
🔨 Welcome to Accentor API

[database]

  pg:console - Open database console
  pg:setup   - Setup postgres in project folder
  pg:start   - Start postgres instance
  pg:stop    - Stop postgres instance

[general commands]

  menu       - prints this menu

```
If we want to, we can customize this message.

The `pg:*` commands allow us to keep the database fully contained within the project folder. Running `pg:start` should be plenty to setup and run a database in `tmp/postgres`. (I adopted this approach so I can easily switch between major versions of postgres accross different projects).

### Github action
@chvp For this action to work after dependabot (and trigger actions), you might have to add a PAT for dependabot to use. [source](https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 # See https://git-scm.com/docs/gitattributes for more about git attribute files.
 
-# Mark the gemset.nix file as having been generated.
+# Mark the lockfiles file as having been generated.
+flake.lock linguist-generated
 gemset.nix linguist-generated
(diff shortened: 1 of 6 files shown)


chvp (approved):
Played around with it locally, everything seems to work.

PR #337, opened 2022-05-21
Don't use bundix' gemset in devshell

chvp:
It makes adding a dependency during local development unnecessarily difficult
and it doesn't really provide that much benefit as far as I can tell.

It's not perfect, since native dependencies in the gems can link to garbage-collected paths, but a `bundle pristine` fixes those.

Note that I don't necessarily need this change. If there is good reason to keep things as-is, that is also fine, but @rien struggled with the "adding-a-local-dependency" thing yesterday (on a different project).

This also moves the postgres data to the `PRJ_DATA_DIR` provided by devshell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,22 +48,33 @@
           default = accentor-api;
           accentor-api = pkgs.devshell.mkShell {
             name = "Accentor API";
-            packages = [
-              gems
-              (pkgs.lowPrio gems.wrappedRuby)
-              pkgs.bundix
-              pkgs.ffmpeg
-              pkgs.nixpkgs-fmt
-              pkgs.postgresql_14
+            imports = [ "${devshell}/extra/language/c.nix" ];
+            packages = with pkgs; [
+              (lowprio binutils)
+              findutils
+              gnumake
+              ruby_3_1
+              bundix
+              ffmpeg
+              nixpkgs-fmt
+              postgresql_14
             ];
             env = [
               {
+                name = "GEM_HOME";
+                eval = "$PRJ_DATA_DIR/bundle/$(ruby -e 'puts RUBY_VERSION')";
+              }
+              {
                 name = "PGDATA";
-                eval = "$PWD/tmp/postgres";
+                eval = "$PRJ_DATA_DIR/postgres";
               }
               {
                 name = "DATABASE_HOST";
                 eval = "$PGDATA";
+              }
+              {
+                name = "PATH";
+                prefix = "$GEM_HOME/bin";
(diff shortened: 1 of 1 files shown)


robbevp:
I've struggled with adding new gems (or simply updating the Gemfile.lock when the ruby version changes). The best thing I have found up till now is the following:
```
$ nix-shell -p ruby_3_1 bundix
[nix-shell]$ bundle lock && bundix
```
But this is not ideal, since it relies on the version of nixpkgs that the user has, rather than the one we pinned the lockfile to.

However, AFAIK including the gemset like this is the only way to get `ruby-vips` to properly work without installing it globally with homebrew or similar. (See [this discussion](https://discourse.nixos.org/t/cant-get-ruby-vips-to-work-in-nix-shell-on-macos/16585)).

robbevp:
I also cannot load this devshell:
```
direnv: loading ~/Code/Accentor/api/.envrc
direnv: using flake
error: builder for '/nix/store/pk0n1z98y1vj5g35qsx8j3m4big9y44v-devshell-dir.drv' failed with exit code 25;
       last 1 log lines:
       > error: collision between `/nix/store/58ww4wrbm3il9cg97wa9zrwv9wn6vwlw-binutils-wrapper-2.35.2/bin/strip' and `/nix/store/hzl1axrp4fvxf31n62x785xhihbvzdz0-gcc-wrapper-11.2.0/bin/strip'
       For full logs, run 'nix log /nix/store/pk0n1z98y1vj5g35qsx8j3m4big9y44v-devshell-dir.drv'.
error: 1 dependencies of derivation '/nix/store/bvgxjhjkidpwld8iif7wcskwl3lrf2ar-Accentor-API-env.drv' failed to build
```

chvp:
Should be fixed for macOS now. How does adding a dependency work with the workaround you propose? The `GEM_PATH` will still be a read-only location in the nix store, no?

robbevp:
This devshell still give the same error as before (I also had to change `lowprio binutils` to `lowPrio binutils` to make it work).

> How does adding a dependency work with the workaround you propose? The `GEM_PATH` will still be a read-only location in the nix store, no? Or do you just add the dependency to the Gemfile and then call `bundle lock` and `bundix` (so not letting bundle do any installation work?)?

The latter. I add extra dependency to the Gemfile and then run the command. `bundle lock` will just update the lockfile (and generate stuff in `vendore/cache` but not actually install the dependency)

chvp:
OK, guess that's a good enough way to add extra dependencies.

Issue #215, opened 2021-05-29
Investigate why Rack can't be bumped


chvp:
I spent the morning trying to bump Rack to 2.2.3, but it seems there are more issues than https://github.com/rack/rack/issues/1619/. Puma seems to get fully stuck when trying to stream a file, even if the workaround mentioned in https://github.com/rack/rack/issues/1619 is applied. I haven't yet been able to figure out why or where it gets stuck.

chvp:
Changes I used to test are in the https://github.com/accentor/api/tree/feature/bump-rack branch

robbevp:
Did you try this patch with incremental updates? Might be good to pinpoint whether this additional issue was present in 2.2.0 or showed up in a later release

chvp:
I did not. This is the same issue why the original 2.2 update was reverted though.

robbevp:
Tried some things with debugger to figure out this issue.
For `audio_with_file`:
Something goes wrong when setting the response_body, with the following code 'setting headers' is never printed to the log.
```rb
  def audio_with_file(path, mimetype)
    Rack::File.new(nil).serving(request, path).tap do |(status, headers, body)|
      logger.debug 'start file'
      self.status = status
      logger.debug 'setting body'
      self.response_body = body

      logger.debug 'setting headers'
      headers.each do |name, value|
        response.headers[name] = value
      end
      logger.debug reponse.headers

      response.content_type = mimetype
      response.headers['accept-ranges'] = 'bytes'
      logger.debug 'finished 
    end
  end
```

Things I've tried:
* Setting the header(s) before setting the reponse body.

For `audio_with_stream`:
If I force it to use this method (without a range specified) it seems to work, even without setting the 'Last-Modified'-header manually.

@chvp I'm not sure why we have to methods to send audio. If `audio_with_stream` works, is there a reason to not use that to serve all audio?

chvp:
`audio_with_file` exists to handle the easy case where we serve directly from the filesystem. `Rack::File` also handles the case where multiple byte ranges are requested. [This](https://www.rubydoc.info/github/rack/rack/Rack/Utils#get_byte_ranges-class_method) exists, but serving multiple ranges isn't easy either (should be a multipart response, IIRC).

I'm not opposed to unifying the streaming responses. As noted in `audio_with_stream`, range requests with multiple ranges don't happen in practice, so I'm fine with not supporting those.

robbevp:
[...]

PR #347, opened 2022-06-20
Fix responses not being streamed

chvp:
The ETag calculation was still happening, so somewhat slow
responses (e.g. transcodes of an hour of music) could take a minute until first
byte sent. This means that clients time out when trying to stream these files.

Related: #215 #216


chvp:
(This is the other end of the fix. The android app now doesn't get stuck on timeouts, this should prevent the timeouts from even happening.)

robbevp (approved):

PR #430, opened 2023-02-20
Make sure we return all genres when filtering tracks by genre

### LINKED ISSUE #429: Getting tracks by genre only includes the requested genre for those tracks
##### Reproduction steps

1. Go to a genre.
2. Notice the tracks that had multiple genres change to only have one genre. Note that this change only happens in the local store (it is not replicated to the backend).

##### Expected behaviour

The genres should not change.

robbevp:
Fixes #429 

### DIFF
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -36,7 +36,7 @@ class Track < ApplicationRecord
   scope :by_filter, ->(filter) { where('"tracks"."normalized_title" LIKE ?', "%#{Track.normalize(filter)}%") }
   scope :by_artist, ->(artist) { joins(:artists).where(artists: { id: artist }) }
   scope :by_album, ->(album) { where(album:) }
-  scope :by_genre, ->(genre) { joins(:genres).where(genres: { id: genre }) }
+  scope :by_genre, ->(genre) { where(id: joins(:genres).where(genres: { id: genre }).select(:id)) }
 
   def merge(other)
     af = other.audio_file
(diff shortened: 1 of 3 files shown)


robbevp:
I've just fix the issue now, but maybe we need to check this for all other scopes, and writes some additional tests as well?

app/models/track.rb
```diff
@@ -36,7 +36,7 @@ class Track < ApplicationRecord
   scope :by_filter, ->(filter) { where('"tracks"."normalized_title" LIKE ?', "%#{Track.normalize(filter)}%") }
   scope :by_artist, ->(artist) { joins(:artists).where(artists: { id: artist }) }
   scope :by_album, ->(album) { where(album:) }
-  scope :by_genre, ->(genre) { joins(:genres).where(genres: { id: genre }) }
+  scope :by_genre, ->(genre) { where(id: joins(:genres).where(genres: { id: genre }).select(:id)) }
```
chvp, line 39: I think it should be a bit more efficient to do this:
```suggestion
  scope :by_genre, ->(genre) { where(id: TrackGenre.where(genre_id: genre).select(:track_id)) }
```
Right now the query that was built up before gets copied to the subquery, which means all conditions would have to be executed twice. It doesn't really matter for our usecases though since AFAIK we only ever filter by one scope in existing frontends.

chvp (approved):
I did some testing myself, it seems like this is only a problem when we use `<relation>_ids`.

app/models/track.rb
robbevp: since we use `has_and_belongs_to_many`, we don't actually have a `TrackGenre` model.
I though the implicit self before `joins(:genres)` refered to the `Track` class and not the query chain. Maybe we just need to explicitly mention the class to escape the query that is being build up?
```
scope :by_genre, ->(genre) { where(id: Track.joins(:genres).where(genres: { id: genre }).select(:id)) }
```

chvp: That would work. (I didn't realize that we used `has_and_belongs_to_many`.)

robbevp:
I though the change above would be a simple change, but it seems to break another test. I'll try to check tonight

chvp (approved):

PR #448, opened 2023-04-22
Upgrade ruby to 3.2.2

robbevp:
Support for ruby 3.2.x landed in nixpkgs/nixos a few weeks ago.

I've not (yet) enabled YJIT in our default devshell, but we might want to do so, if we end up using this in production.


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '~> 3.1.0'
+ruby '~> 3.2.2'
```
robbevp, line 4: Maybe we should not require an exact version here, but use `~> 3.2`?

chvp: `~> 3.2.2` is already not an exact version. It means `>= 3.2.2 && < 3.3`, which seems like the specifity we want for Ruby versions, since the minor version number is a new language level that requires more care.

robbevp: ah, yes. nvm

chvp:
There's a lot of VIPS warnings/messages in the test run. Any idea what that's about?

robbevp:
> There's a lot of VIPS warnings/messages in the test run. Any idea what that's about?

I've seen this in other projects as well - this is unrelated to ruby 3.2 (the warnings are also present on the [last actions on main](https://github.com/accentor/api/actions/runs/4772476492/jobs/8485007873)) and seem to only happen in CI (I've not seen this warning locally or in production for any of the other projects that consistently have these warnings in github actions).

chvp (changes requested):
https://github.com/accentor/api/blob/main/.rubocop.yml#L14 should also be updated.

robbevp:
> https://github.com/accentor/api/blob/main/.rubocop.yml#L14 should also be updated.

Updated and rebased

chvp (approved):

PR #471, opened 2023-09-02
Fix error if tags contain null byte(s)

chvp:
No test, because I couldn't figure out how to create a file with such tags. I did test locally that the problem files I had scanned without issue.

### DIFF
--- a/app/models/rescan_runner.rb
+++ b/app/models/rescan_runner.rb
@@ -86,14 +86,14 @@ class RescanRunner < ApplicationRecord
     return if AudioFile.exists?(location:, filename: relative_path.to_s)
 
     tag = WahWah.open(path)
-    t_artist = tag.artist&.unicode_normalize
-    t_albumartist = tag.albumartist.present? ? tag.albumartist&.unicode_normalize : t_artist
-    t_composer = tag.composer&.unicode_normalize
-    t_title = tag.title&.unicode_normalize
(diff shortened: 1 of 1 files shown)


app/models/rescan_runner.rb
```diff
@@ -204,4 +204,12 @@ def convert_year(tag)
   rescue Date::Error
     Date.new(0)
   end
+
+  def clean_string(s)
+    # ID3v2 tags sometimes include null bytes to "split" fields. It doesn't
+    # occur very often though, so remove null bytes instead of trying to parse
+    # that. Database errors out on null bytes, so we can't just do nothing.
```
robbevp, line 211:
```suggestion
```suggestion
    # ID3v2 tags sometimes include null bytes to "split" fields. It doesn't
    # occur very often, so we simply remove null bytes instead of
    # trying to parse them. Otherwise our database will throw an error, 
    # since null bytes aren't allowed in strings.
```

robbevp (commented):
I've given a suggestion for a reworded comment.

While I think this fix is fine, I do wonder if it doesn't make more sense to upstream this fix to `WahWah`? Is there a scenario in which someone would want to keep the null bytes?

chvp:
I don't think upstreaming is the right fix. As mentioned in the comment, some programs use the null bytes to separate multiple values in a tag. Other consumers of WahWah might want to go through the effort of using that information.

robbevp (approved):

PR #478, opened 2023-09-09
Switch active job queue adapter

### LINKED ISSUE #473: Switch active job backend to `GoodJob`
[`GoodJob`](https://github.com/bensheldon/good_job) offers a few advantages over DelayedJob:
* Uses LISTEN/NOTIFY instead of polling
* Multithreaded instead of single threaded
* Built-in support for CRON jobs
* Built-in job dashboard (if we want that)

The only downside is that we need to go through Active Job and can no longer use inline scheduling.

## Migration plan
* [ ] Add GoodJob and convert background tasks to use Active Job
* [ ] Deploy a version to production with both DelayedJob and GoodJob
* [ ] Remove DelayedJob

See: https://github.com/bensheldon/good_job#migrate-to-goodjob-from-a-different-activejob-backend

robbevp:
Fixes #473 

This PR switches the active job backend from DelayedJob to [GoodJob](https://github.com/bensheldon/good_job#migrate-to-goodjob-from-a-different-activejob-backend). DelayedJob will be removed later, since we don't have a migration planned for the jobs that are currently queued in a production environment.

I've created explicit job classes for all things which currently got scheduled with `delay()`. Since these classes only exist to call a method on the model, I didn't add tests for these. (I can add them, but I wouldn't want to test the actual model logic there, so they would simply test whether a method exists).

In some cases I think it could make sense to move more code to the job class, but I haven't done this for now. (For `CalculateContentLengthJob` and `ConvertTranscodeJob` most - if not all - of the model code could probably also exist in the job class.)

My last two commits are optional and can be removed if we don't want those. [State before those two commits](https://github.com/accentor/api/tree/018df94292010746f4353a28a227ccb7e5737878)
My reasoning for them:
### `035f7b3` - Queue structure
I renamed/reorganised these queue to time based queues. We switch to a fixed number of queues, based on the maximum time a job should allowed to be enqueued for, before being picked up by a worker. This makes the ranking and/or priority a lot more explicit, without going done to minute details like priority per queue leads to. The exact numbers don't matter that much (and could be changed), but I went for:
* within_30_seconds
* within_5_minutes
* within_30_minutes
* within_24_hours

A deployment can then configure their workers, so that none of these queues ever overflow and don't respect their time limits. (This can be done with different workers, or with a thread limit within one worker, see: https://github.com/bensheldon/good_job#optimize-queues-threads-and-processes)
This approach is based on Nate Berkopec's advice about running Sidekiq, but also applies here.

### `fd74c89` - Mount engine
Good job includes a decent dashboard to monitor the jobs and processes/workers. I use this dashboard quite a bit at work (to monitor a bit and to retry failed jobs when relevant), but I don't know if we would want this in accentor.
I've currently mounted it at `/good_job`, but wasn't sure if we would want that. 
Two considerations:
* We need to re-add some middleware to make this work
* We need to roll some kind of authentication. Maybe basic auth with nginx is good enough here?

We could also make this an option (either in rails or in the nixos module), but I'd hesitate adding the complexity if no-one ends up using this.


test/controllers/tracks_controller_test.rb
```diff
@@ -282,6 +282,7 @@ class TracksControllerAudioTest < ActionDispatch::IntegrationTest
     location = Location.create(path: Rails.root.join('test/files'))
     flac = Codec.create(mimetype: 'audio/flac', extension: 'flac')
     audio_file = create(:audio_file, location:, filename: '/base.flac', codec: flac)
+    perform_enqueued_jobs
```
robbevp, line 285: Most tests will have the jobs executes inline, but Rails sometimes overrides the queue_adapter. We could work around this, but we can just as well force the jobs to execute.
See: https://github.com/rails/rails/issues/37270

test/controllers/tracks_controller_test.rb
```diff
@@ -303,18 +304,14 @@ class TracksControllerAudioTest < ActionDispatch::IntegrationTest
     flac = Codec.create(mimetype: 'audio/flac', extension: 'flac')
     audio_file = create(:audio_file, location:, filename: '/base.flac', codec: flac)
     track = create(:track, audio_file:)
-    get audio_track_url(track, codec_conversion_id: codec_conversion.id)
-    File.delete(TranscodedItem.last.path)
+    transcode = TranscodedItem.create(audio_file:, codec_conversion:)
```
robbevp, line 307: This test failed before, because of the rather cryptic `IOError: not opened for reading` exception.

Creating the transcode without going through a request fixed this issue. IMO this also makes the intention of the test easier to read/understand.

Backtrace:
```
IOError: not opened for reading
    app/models/transcoded_item.rb:33:in `read'
    app/models/transcoded_item.rb:33:in `readpartial'
    app/models/transcoded_item.rb:33:in `copy_stream'
    app/models/transcoded_item.rb:33:in `block in do_conversion'
    app/models/transcoded_item.rb:33:in `open'
    app/models/transcoded_item.rb:33:in `do_conversion'
 ```

**Comment by robbevp** [comment]
Fixed the failing test and updated my comment to refer to the right commits

File: test/controllers/tracks_controller_test.rb (hunk 2 above)
Comment (chvp) [reply test/controllers/tracks_controller_test.rb:307]: Seeing as this is part of the `tracks_controller_test`, I don't entirely agree with not going through a request.

**Comment by chvp** [comment]
I'd not add the dashboard. As you mentioned, it needs auth and a bunch of middleware we currently don't have. If at any point in the future we notice missing the dashboard we can still add it.

File: test/controllers/tracks_controller_test.rb (hunk 2 above)
Comment (robbevp) [reply test/controllers/tracks_controller_test.rb:307]: The actual request that we want to test is on L313. This request was in no way checked or asserted, so I assumed this was part of the setup.
You wrote these tests, so let me know if I interpreted that wrong

**Comment by robbevp** [comment]
> I'd not add the dashboard. As you mentioned, it needs auth and a bunch of middleware we currently don't have. If at any point in the future we notice missing the dashboard we can still add it.

Dropped this commit

File: test/controllers/tracks_controller_test.rb (hunk 2 above)
Comment (chvp) [reply test/controllers/tracks_controller_test.rb:307]: Ah, you're right. Hadn't looked at it in detail, my bad.

File: app/models/content_length.rb [hunk 3]
```diff
@@ -22,7 +22,7 @@ def self.destroy_all_and_recalculate
       next unless Rails.application.config.recalculate_content_length_if.call af
 
       CodecConversion.find_each do |cc|
-        af.delay(queue: :content_lengths_backlog).calc_audio_length(cc)
+        CalculateContentLengthJob.set(queue: :within_30_minutes).perform_later(af, cc)
```
chvp, line 25: It's entirely infeasible to recalculate all content lengths (that qualify) within 30 minutes. This happens very rarely, but if it does it takes at least a few days. Is it possible to create a "no deadline" queue where these jobs can be scheduled to?

chvp (commented):
Can you create a matching PR on the `flake` repo to change the way the runner is executed? Or I can do it as well, if you don't want to.

app/models/content_length.rb
robbevp: Good point, I've renamed `:within_24_hours` to `:whenever` (as I don't think we need the distinction between those two, and the latter is more explicit).

Due to this change, I've put the `TranscodeCacheCleanJob` in `:within_30_minutes` as it should be possible for this job to be picked up rather quickly.

app/jobs/transcode_cache_clean_job.rb
```diff
@@ -1,5 +1,5 @@
 class TranscodeCacheCleanJob < ApplicationJob
-  queue_as :transcode_cache_cleaner
+  queue_as :within_30_minuts
```
chvp, line 2:
```suggestion
  queue_as :within_30_minutes
```

chvp (commented):
Since (as far as I can tell at least), the priority of the queues is not defined in-repo anymore, we should at least mention them in the setup instructions I think?

robbevp:
> Since (as far as I can tell at least), the priority of the queues is not defined in-repo anymore, we should at least mention them in the setup instructions I think?

Good suggestion. We didn't have anything about starting/managing background workers, but I've added a section.

Note: I've also changed Rails' default queue to be `within_5_minutes` (AFAIK this should only be `ActiveStorage::AnalyseJob`and `ActiveStorage::PurgeJob` for us - if need be, we can specify this per job in the config [docs](https://guides.rubyonrails.org/configuring.html#config-active-storage-queues-analysis))

chvp (approved):

PR #487, opened 2023-10-01
Add download action for tracks

chvp:

### DIFF
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -1,7 +1,7 @@
 class TracksController < ApplicationController
   include ActionController::Live
 
-  before_action :set_track, only: %i[show update destroy audio merge]
+  before_action :set_track, only: %i[show update destroy audio download merge]
 
   has_scope :by_filter, as: 'filter'
   has_scope :by_album, as: 'album_id'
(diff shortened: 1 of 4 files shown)


app/controllers/tracks_controller.rb
```diff
@@ -78,6 +78,13 @@ def audio
     end
   end
 
+  def download
+    audio_file = @track.audio_file
+    raise ActiveRecord::RecordNotFound.new('track has no audio', 'audio') unless audio_file.present? && audio_file.check_self
+
+    send_file audio_file.full_path
```
robbevp, line 85: I'd add:
```suggestion
    send_file audio_file.full_path, filename: audio_file.filename, disposition: :attachment
```

robbevp: Maybe even edit the filename so that it contains the codec's extension? Systems can behave differently when files contain the correct mime headers but no extension.

robbevp (commented):
Looks okay - but I'd add an explicit filename to avoid downloading a file with the id as filename.

app/controllers/tracks_controller.rb
chvp: According to [the docs](https://api.rubyonrails.org/classes/ActionController/DataStreaming.html#method-i-send_file), the default `:filename` is the basename of the file on the file system. `disposition: "attachment" ` is also the default.

robbevp: Sorry, I was thinking of `send_blob`

robbevp (approved):

PR #491, opened 2023-10-14
Update to rails 7.1

chvp:
Still need to go through the new framework defaults file and make sure we can enable all of them.


robbevp (changes requested):
I would keep the upgrade to rails 7.1 separate from adding `rubocop-facotry_bot` to the gems and linting setup (even though I'm in favor of adding that gem 🙂)

chvp:
I've made a PR in #494 so that I can rebase this when that one is merged instead of factoring out the changes.

config/application.rb
```diff
@@ -17,7 +17,15 @@
 module Accentor
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 7.0
+    config.load_defaults 7.1
+
+    # Use the latest cache format version. We use the memory store, so no issues with upgrades
+    config.active_support.cache_format_version = 7.1
+
```
robbevp, line 24: Isn't this also set by `config.load_defaults 7.1`?

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,8 @@
 # Be sure to restart your server when you modify this file.
 
-# Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += %i[passw secret]
+# Configure parameters to be partially matched (e.g. passw matches password) and filtered from the log file.
+# Use this to limit dissemination of sensitive information.
+# See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
+Rails.application.config.filter_parameters += %i[
+  passw secret token _key crypt salt certificate otp ssn
+]
```
robbevp, line 8: When upgrading to 6.1 we opted to not include these changes, since we don't have these kind of parameters: https://github.com/accentor/api/pull/156#discussion_r541586694

robbevp (changes requested):

robbevp (approved):

PR #498, opened 2023-10-27
Move recalculation of content lengths to background job

### LINKED ISSUE #497: ffmpeg:check_version takes a very long time if the version changed
In #478, the scheduling of the jobs when the version changed was made inline instead of being done from a job. This means that all those jobs need to be scheduled before the server starts (at least with the way it's configured on our deployment).

robbevp:
Fixes #497 

I moved the iteration over all audio files to its own job. We'd want to run `ContentLength.destroy_all` ASAP after boot, since a version mismatch could result in issues for users. If this turns out to still be a bottleneck, we can also switch to `ContentLength.delete_all` (as we don't have any lifecycle callbacks for content lengths, or need to delete associated models when destroying).


test/jobs/recalculate_content_lengths_job_test.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    # rubocop:disable Rails/SkipsModelValidations
+    # We want to avoid our own before_save callbacks to manually set the audio_file length and track age
+    @audio_file.update_column(:length, 1)
+    @audio_file.track.update_column(:created_at, 1.day.ago)
+    # rubocop:enable Rails/SkipsModelValidations
+
+    assert_enqueued_jobs 1 do
+      RecalculateContentLengthsJob.perform_now
+    end
```
robbevp, line 57: I'm not sure why this fails and L39:41 doesn't. When debugging I can see that a job is created and the same code path is taken.

robbevp: NVM, I was looking at the wrong line number 🤦

test/jobs/recalculate_content_lengths_job_test.rb
```diff
@@ -0,0 +1,59 @@
...
+    # rubocop:enable Rails/SkipsModelValidations
+
+    assert_no_enqueued_jobs do
+      RecalculateContentLengthsJob.perform_now
+    end
+  end
+
+  test 'should enqueue job if audio is longer than config' do
+    io = StringIO.new Rails.root.join('test/files/base.flac').read
+    AudioFile.any_instance.stubs(:convert).returns(io)
```
chvp, line 30: I'm a bit confused on why this is both in the `setup` and (some) `test`s.

chvp (commented):
Just some questions on the tests, but I see that the same thing was also in the previous tests, so maybe it's not necessary to fix it here?

robbevp:
Not sure how that got there.
Anyway, I cleaned up the tests so that all shared setup is only present in the setup.

test/jobs/recalculate_content_lengths_job_test.rb
```diff
@@ -0,0 +1,50 @@
...
+    @audio_file = create(:audio_file)
+    @codec_conversion = create(:codec_conversion)
+    @track = create(:track, audio_file: @audio_file)
+
+    ContentLength.destroy_all
+  end
+
+  test 'should not enqueue job if audio is longer than config and track is older than config' do
```
chvp, line 15:
```suggestion
  test 'should not enqueue job if audio is shorter than config and track is older than config' do
```

chvp (approved):

PR #504, opened 2023-11-08
Make sure that content-length header can never be deleted

### LINKED ISSUE #503: `content-length` header isn't set for streaming audio responses anymore since rails 7.1
At least the android client, but probably also others need the header to properly cache responses.

##### Reproduction steps

1. Fetch transcoded audio before the rails 7.1 update and observe that the header is there.
2. Fetch transcoded audio after the rails 7.1 update and observe that the header is not there.

##### Expected behaviour

The header should be there so clients can properly cache the response.

chvp:
Once the response is `commit`ted, the headers can't be edited anymore, which is exactly what we want in this case.

Unfortunately, this can't be added to the tests, because the way responses and `ActionController::Live` work in tests just doesn't allow it (in fact, the tests already "test" for the presence of this header, and the tests still succeed even though in production the header is not present).

Fixes #503.

### DIFF
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -116,6 +116,10 @@ class TracksController < ApplicationController
     response.content_type = mimetype
     response.headers['accept-ranges'] = 'bytes'
     response.headers['content-length'] = (last_byte - first_byte + 1).to_s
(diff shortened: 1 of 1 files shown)


chvp:
(I've already applied this fix in our production instance, because it was quite annoying.)

robbevp (approved):

PR #545, opened 2024-02-25
Enhc/add individual playlist items

robbevp:
Adding a single playlist item to an existing playlist through `PlaylistsController#update` takes longer and longer, the more items there are (since we destroy and create all items). This PR adds an additional endpoint to simply create a playlist item.

Two choices I made:
* A user is not able to specify the order. The item always gets added as the last item in the list
* The user (but more realistically: the client the user is using) does need to specify the type of item, to check that it matches the playlist's type.


robbevp:
Updated after some discussion with @chvp 

We have an `PlaylistsController#add_item` action instead of a separate `PlaylistItemsController` with just a create action

chvp (approved):

PR #622, opened 2024-08-17
Add optional `application` attribute to auth tokens

robbevp:
While working on a separate frontend app, I noticed that we use the `user_agent` attribute a bit inconsistent. For browser apps we track the actual user agent, but for native apps we use this to track the device info and the app info. 
A user can't currently distinguish two browser apps in their auth tokens, since they both just show the user agent string.

This PR allows us to separately track the application and the device/OS that is used for the auth token. I'd propose that:
* `application` is for the app and its version. Eg. `Accentor 0.14.1 on android`, `web 0.1.2`, ... 
* `user_agent` is for the device/OS/browser used. For browser apps this can remain the same, for native apps it could be something like `macOS 14.5 (Build 23F79)` or `Android 13 on Fairphone 4`
For both of these, I wouldn't define a fixed format - as long as the right info is in the right field

I've made this attribute optional, so we can slowly roll out support in various apps. Maybe in time we could make this required?

## Two questions/points of discussion:
* `application` feels a bit generic, but I couldn't come up with anything specific
* I currently get this info from a param. I'd also be ok with defining a specific header to be used (or a combination of both, like for user agent).

I've not added a test, since our controller tests don't check the attributes of the objects that are created.

### DIFF
--- a/app/controllers/auth_tokens_controller.rb
+++ b/app/controllers/auth_tokens_controller.rb
@@ -25,7 +25,7 @@ class AuthTokensController < ApplicationController
     end
 
     @auth_token = AuthToken.new(
-      { user_agent: request.headers[:'user-agent'] }
+      { user_agent: request.headers[:'user-agent'], application: params[:application] }
           .merge(params[:auth_token].present? ? permitted_attributes(AuthToken) : {})
           .merge(user:)
     )
--- a/app/controllers/auth_tokens_controller.rb
+++ b/app/controllers/auth_tokens_controller.rb
@@ -25,7 +25,7 @@ class AuthTokensController < ApplicationController
     end
 
     @auth_token = AuthToken.new(
-      { user_agent: request.headers[:'user-agent'], application: params[:application] }
+      { user_agent: request.headers[:'user-agent'] }
           .merge(params[:auth_token].present? ? permitted_attributes(AuthToken) : {})
           .merge(user:)
     )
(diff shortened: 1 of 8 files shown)


chvp:
Another idea I had today: allowing users to set a device name on their auth tokens, similar to e.g. matrix. Then they can write "web on work laptop", "game on phone", ... Since the main purpose for this info is identification of sessions for deauthing them, this seems like a generic way to do so (and common in other projects).

robbevp:
> Another idea I had today: allowing users to set a device name on their auth tokens, similar to e.g. matrix. Then they can write "web on work laptop", "game on phone", ... Since the main purpose for this info is identification of sessions for deauthing them, this seems like a generic way to do so (and common in other projects). WDYT?

I don't think i see much benefit in this. I wouldn't want to burden the user with requiring them to enter some extra field with their log in info, so we would still need a good default. (But if there is a good default, I don't think I would ever spent the effort to rename my sessions, ...)

But if we want a default that identifies the app and the device used, we still need a system for how to name the user agent to contain both, or split them up. I thinking making this two separate fields will make it easier to present this info in a consistent way

app/controllers/auth_tokens_controller.rb
```diff
@@ -25,7 +25,7 @@ def create
     end
 
     @auth_token = AuthToken.new(
-      { user_agent: request.headers[:'user-agent'] }
+      { user_agent: request.headers[:'user-agent'], application: params[:application] }
```
chvp, line 28: This shouldn't be necessary, I think? This should/could be part of the permitted_attributes below?

robbevp: no clue why i added this here 🤦

chvp (approved):

PR #644, opened 2024-10-04
Precache actual files instead of storing lengths

chvp:


robbevp (commented):
I went through this a first time. My biggest worry is the performance when deleting a large amount of `TranscodeItem`s.

Some other comments aren't on changes made now, but are closely related to the code we are touching, so it might be a good moment to tackle those as well

app/models/transcoded_item.rb
```diff
@@ -3,44 +3,29 @@
 # Table name: transcoded_items
 #
 #  id                  :bigint           not null, primary key
-#  last_used           :datetime         not null
 #  path                :string           not null
```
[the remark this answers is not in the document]
chvp: Already the case: https://github.com/accentor/api/blob/main/db/schema.rb#L348

app/models/transcoded_item.rb
```diff
@@ -3,44 +3,29 @@
...
-    FileUtils.mv(tmppath, path)
-  end
-
-  private
-
-  def random_path
+  def self.random_path
     uuid = SecureRandom.uuid
     File.join(BASE_PATH, uuid[0..1], uuid[2..3], uuid)
   end
```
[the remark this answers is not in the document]
chvp: I have no strong opinion on this personally. I guess it would make moving the base path easier if this is not in the DB.

app/jobs/queue_transcoded_items_for_codec_conversion_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class QueueTranscodedItemsForCodecConversionJob < ApplicationJob
+  queue_as :within_30_seconds
+
+  def perform(codec_conversion)
+    AudioFile.find_each do |af|
+      TranscodedItem.find_by(audio_file: af, codec_conversion:)&.destroy
```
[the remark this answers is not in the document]
chvp: I mean, that's why this is in a job, right? But you raise some good ideas on making this more performant, will implement.

app/models/codec_conversion.rb
```diff
@@ -10,20 +10,16 @@
 
 class CodecConversion < ApplicationRecord
   belongs_to :resulting_codec, class_name: 'Codec'
-  has_many :content_lengths, dependent: :destroy
   has_many :transcoded_items, dependent: :destroy
```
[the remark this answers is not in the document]
chvp: Hmm, you're right. If implement the suggestion from above about scoping transcoded_item paths by codec conversion then we can do a `rm_f` and a `delete_all` in a `before_destroy`.

app/jobs/create_transcoded_item_job.rb
```diff
@@ -0,0 +1,21 @@
+class CreateTranscodedItemJob < ApplicationJob
+  queue_as :within_30_seconds
+
+  def perform(audio_file, codec_conversion)
+    # Check that this transcoded_item was not created while this job was in the queue
+    return if TranscodedItem.find_by(audio_file:, codec_conversion:).present?
+
+    path = TranscodedItem.random_path
+    FileUtils.mkdir_p Pathname.new(path).parent
+    audio_file.convert(codec_conversion, path)
```
robbevp, line 10: Since we have a unique index on these paths, this could potentially:
* return a uuid that was already used for track A
* override the output of the existing transcode A with transcode B
* fail to save transcode B because of the unique index
* create a lot of confusion, since transcode A now returns audio for B

Maybe `TranscodedItem.random_path` should return a path that was not yet used?

app/jobs/create_transcoded_item_job.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    TranscodedItem.transaction do
+      # Check that the transcoded item was not created while we were executing
+      if TranscodedItem.find_by(audio_file:, codec_conversion:).present?
+        FileUtils.rm_f path
+      else
+        TranscodedItem.create(audio_file:, codec_conversion:, path:)
```
robbevp, line 17: I prefer to use `create!` in these situations. If we don't expect the create/update/destroy to fail and/or don't handle the boolean that is returned, we should raise an error so we know about the failure and can improve the code

robbevp (commented):
Two extra things I realized yesterday after giving my first review

app/jobs/create_transcoded_item_job.rb
chvp: Good point.

app/models/transcoded_item.rb
```diff
@@ -3,44 +3,48 @@
...
-  def do_conversion
-    tmppath = random_path
-    FileUtils.mkdir_p Pathname.new(tmppath).parent
-    File.open(tmppath, 'w') { |f| IO.copy_stream(audio_file.convert(codec_conversion), f) }
-    FileUtils.mkdir_p Pathname.new(path).parent
-    FileUtils.mv(tmppath, path)
+  def self.uuid_for(codec_conversion)
+    loop do
+      uuid = SecureRandom.uuid
+      return uuid if TranscodedItem.find_by(codec_conversion:, uuid:).nil?
```
robbevp, line 34:
```suggestion
      return uuid unless TranscodedItem.exists?(codec_conversion:, uuid:)
```

test/jobs/queue_transcoded_items_for_codec_conversion_job_test.rb
```diff
@@ -0,0 +1,4 @@
+require 'test_helper'
+
+class QueueTranscodedItemsForCodecConversionJobTest < ActiveJob::TestCase
+end
```
robbevp, line 4: This is still missing tests.

I would test:
* Are the files deleted?
* Are the expected amount of jobs created?

app/models/codec_conversion.rb
```diff
@@ -10,20 +10,27 @@
...
 
-  def queue_content_length_calculations
-    ContentLength.where(codec_conversion: self).destroy_all
-    AudioFile.find_each do |af|
-      CalculateContentLengthJob.perform_later(af, self)
-    end
+  def delete_transcoded_items
+    transcoded_items.delete_all
+    FileUtils.rm_rf TranscodedItem.codec_conversion_base_path(self)
   end
```
robbevp, line 35: I'd add a model test to verify this logic

robbevp (changes requested):

app/controllers/tracks_controller.rb
```diff
@@ -57,22 +57,15 @@ def audio
...
-        # AudioFile will only do the conversion if the `ContentLength` doesn't exist yet.
-        content_length = audio_file.calc_audio_length(conversion)
-        audio_with_stream(audio_file.convert(conversion), conversion.resulting_codec.mimetype, content_length.length)
+      transcoded_item = TranscodedItem.find_by(audio_file:, codec_conversion: conversion)
+      unless transcoded_item.present? && File.exist?(transcoded_item.path)
+        transcoded_item.destroy if transcoded_item.present? # The file was lost. This shouldn't happen, so delete the item
+
+        # This does the conversion inline
+        CreateTranscodedItemJob.perform_now(audio_file, conversion)
+        transcoded_item = TranscodedItem.find_by(audio_file:, codec_conversion: conversion)
```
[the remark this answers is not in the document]
chvp: This leaves us open to edge cases where the `TranscodedItem` is not actually created by the `perform_now` (e.g. when a job was also scheduled before the request and finishes during the transcode of the `perform_now`.

robbevp (changes requested):
The code changes are all ok to me. I realised just now that we still need to update our readme, since it mentions `RAILS_TRANSCODE_CACHE`

robbevp (approved):

PR #647, opened 2024-10-05
Remove obsolete `check_file_attributes` method

chvp:
This was only used in a migration, and new deployments will not actually need it.

### DIFF
--- a/app/models/audio_file.rb
+++ b/app/models/audio_file.rb
@@ -36,13 +36,6 @@ class AudioFile < ApplicationRecord
     false
   end
 
-  def check_file_attributes
-    return unless check_self
-
-    tag = WahWah.open(full_path)
-    update(length: tag.duration, bitrate: tag.bitrate || 0, sample_rate: tag.sample_rate || 0, bit_depth: tag.bit_depth || 0)
-  end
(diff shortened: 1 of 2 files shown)


robbevp (commented):
Not sure I agree with deleting this, since the particular migration still references this function.

Three options:
* We leave this here
* We place the content of the method inside the migration
* We remove old migrations (since new setups should use `db:schema:load` and we don't support people who haven't updated in 4 years).

I think I prefer option 2 in this case. I know some people who are in favor of the last option but I'm not fully convinced

chvp:
AFAIK, it is commonly understood that running old migrations will not work in most rails apps. I definitely don't want to go for option 1. Leaving obsolete code in the codebase is just confusing and unnecessary.

I would rather remove the old migrations TBH.

robbevp:
Sure. I just had a look at the migration in question, and it still uses delayed job which wouldn't work anyway. How about we delete all migrations that are more than 2 years old?
We do need an exception for migrations from rails and good job, as their generators could re-create a migration that we already deleted (but with a different timestamp)

robbevp (approved):

Issue #297, opened 2022-02-12
Flaky test in trackscontroller


chvp:
Every so often CI fails the following test:

```
Error:
TracksControllerAudioTest#test_should_not_create_transcoded_item_if_it_already_exists_but_should_queue_if_file_is_gone:
RuntimeError: Wrapped undumpable exception for: ActiveRecord::NotNullViolation: PG::NotNullViolation: ERROR:  null value in column "codec_id" of relation "audio_files" violates not-null constraint
DETAIL:  Failing row contains (18, 10, null, /base.flac, 871610, 984, 8391, 0).

    test/controllers/tracks_controller_test.rb:275:in `block in <class:TracksControllerAudioTest>'

rails test test/controllers/tracks_controller_test.rb:269
```

chvp:
Pretty sure I fixed this in the Rails 7 update.

PR #656, opened 2024-10-10
Make `audio_file#convert` raise an exception on failure

### LINKED ISSUE #651: Add test that checks that `AudioFile#convert` doesn't return before the actual conversion is finished

### LINKED ISSUE #654: Make sure `AudioFile#convert` throws an exception if the underlying ffmpeg process didn't exit with status 0

chvp:
Fixes #654.
Fixes #651.

### DIFF
--- a/test/models/audio_file_test.rb
+++ b/test/models/audio_file_test.rb
@@ -15,6 +15,15 @@
 require 'test_helper'
 
 class AudioFileTest < ActiveSupport::TestCase
+  class ProcessStub
+    def initialize(exit_status)
+      @exit_status = exit_status
+    end
+
+    def success? = @exit_status.zero?
+    def to_i = @exit_status
+  end
+
   def setup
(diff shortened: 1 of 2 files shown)


test/models/audio_file_test.rb
```diff
@@ -15,6 +15,16 @@
 require 'test_helper'
 
 class AudioFileTest < ActiveSupport::TestCase
+  class ProcessStub
+    def initialize(success, exit_status)
+      @success = success
+      @exit_status = exit_status
+    end
+
+    def success? = @success
```
robbevp, line 24: Couldn't we simplify this to:
```suggestion
    def initialize(exit_status)
      @exit_status = exit_status
    end

    def success? = @exit_status == 0
```

robbevp (commented):
One suggestion to simplify the test stub.

The tests are also failing, but this seems to be #297 - so running them again might fix this

robbevp (approved):

PR #657, opened 2024-10-10
Enqueue codec_conversion transcodes on slower queue

### LINKED ISSUE #653: Tweak `QueueTranscodedItemsForCodecConversionJob` so that `CreateTranscodeJob`s are queued on a low-priority queue
This should prevent the `CreateTranscodeJob`s from cannibalizing all progress on other jobs.

chvp:
Fixes #653.


robbevp (commented):
I'd even move this to our slowest queue. If we see these queues as a contrqct wehere the task _should_ be picked up in this time. This will probably not happen when we enqueue these tasks in bulk.

chvp:
In practice there is little difference; both queues are not used by any other jobs. But agree that the semantics of `whenever` are clearer for this case.

robbevp (approved):

PR #659, opened 2024-10-10
Transcode to a temporary file which is only moved when finished

### LINKED ISSUE #652: Tweak `CreateTranscodeJob` to transcode to a temporary file and only move to the final location when finished
This makes sure that no half-transcoded files can end up in the final storage location (in combination with #654).

chvp:
Fixes #652.

### DIFF
--- a/app/models/audio_file.rb
+++ b/app/models/audio_file.rb
@@ -37,6 +37,14 @@ class AudioFile < ApplicationRecord
     false
   end
 
+  def convert_with_tmpfile(codec_conversion, out_file_name)
+    tmp_path = File.join(Dir.tmpdir, "accentor_transcode-#{id}-#{codec_conversion.id}-#{SecureRandom.uuid}")
+    convert(codec_conversion, tmp_path)
+    FileUtils.mv tmp_path, out_file_name
+  ensure
+    FileUtils.rm_f tmp_path
+  end
+
   def convert(codec_conversion, out_file_name)
     parameters = codec_conversion.ffmpeg_params.split
     _stdout, status = Open3.capture2(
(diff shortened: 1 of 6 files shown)


chvp:
@robbevp `Tempfile.create` could not be used because `ffmpeg` then complains about the file already existing. So I went for this alternative approach. (Still need to add tests, but that will be for saturday.)

robbevp:
I agree with the approach. I mostly pushed for `Tempfile.create` so we can rely on the OS to clean up files in case of issues, but `Dir.tmpdir` has the same benefit

test/jobs/create_transcoded_item_job_test.rb
```diff
@@ -34,11 +34,12 @@ class CreateTranscodedItemJobTest < ActiveJob::TestCase
...
     audio_file = @audio_file
     codec_conversion = @codec_conversion
-    AudioFile.define_method :convert, lambda { |_codec_conversion, _out_file_name|
+    AudioFile.define_method :convert, lambda { |_codec_conversion, out_file_name|
+      FileUtils.touch out_file_name
       TranscodedItem.create!(audio_file:, codec_conversion:, uuid: '0000-0000-0000')
     }
 
-    FileUtils.stubs(:rm_f).once
+    FileUtils.stubs(:rm_f).twice
```
robbevp, line 42: To help ourselves in the future, i would add a comment to explain what these two calls are

app/models/audio_file.rb
```diff
@@ -37,6 +37,14 @@ def check_self
     false
   end
 
+  def convert_and_move(codec_conversion, out_file_name)
```
robbevp, line 40: I think we could improve this naming. `convert_and_move` doesn't clearly distinguish this from `convert`, as they both take the same arguments and the end-result is the same. Maybe something like `convert_with_tmpfile`?

robbevp (approved):

PR #680, opened 2024-11-09
Update to rails 8

chvp:


chvp:
`active_model_serializers` doesn't seem to be compatible with Rails 8. It's not maintained at all, so I'll switch away from it. (But I'll do that in a separate MR.)

chvp:
Based on #681, that should be looked at and merged first.

config/puma.rb
```diff
@@ -19,16 +27,12 @@
...
-state_path ENV.fetch('STATEPATH') { 'tmp/pids/server.state' }
-bind ENV.fetch('SOCKETFILE') { 'unix://tmp/sockets/puma.sock' }
 
-# Allow puma to be restarted by `rails restart` command.
+# Allow puma to be restarted by `bin/rails restart` command.
 plugin :tmp_restart
+
+# Specify the PID file. Defaults to tmp/pids/server.pid in development.
+# In other environments, only set the PID file if requested.
+pidfile ENV['PIDFILE'] if ENV['PIDFILE']
```
robbevp, line 38: We lost the config for state_path and socketfile config here, but I think we want to keep these?

robbevp (changes requested):

config/puma.rb
chvp: Good catch, fixed.

robbevp (approved):

PR #681, opened 2024-11-09
Replace AMS with jbuilder

chvp:


robbevp (commented):
I'm not really a fan of how this approach creates two files for every object that we serialize. Two suggestions that we could maybe try out:
1. Did you look into other gems that provide serialization? I've also use blueprinter in projects, and liked their approach.
1. Can we just skip jbuilder and place the tiny bit of logic in the controller? Something like `render json: @albums.map { |a| transform...(album) }`
If that works, we might also be able to place those transform function as a private method in the controller

It might also be worthwhile do benchmark the performamce of these approaches, since that could also influence our decision

chvp:
You're right that we can just get rid of jbuilder as well. We don't have any complex use cases here, so I wouldn't introduce another dependency. Didn't do any performance benchmarking, but I can't imagine that this approach can be slower than doing this through a dependency.

robbevp (approved):

PR #682, opened 2024-11-13
Refactor sending audio in TracksController by switching back to Rack::Files

### LINKED ISSUE #649: Refactor `TracksController#audio_with_file`
After #644, we no longer need `TracksController#audio_with_stream`. This means that we can probably simplify `TracksController#audio_with_file`. One obvious thing that comes to mind is seeking in the file for the requested byte range instead of skipping bytes by reading the stream.

chvp:
Since we never do live transcoding anymore, we can simplify the way we send
audio data to clients a lot. We can remove the usage of `ActionController::Live`
and don't have to do any manual copying to and from streams, all by just
switching back to Rack::Files.

Fixes #649.

### DIFF
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -1,6 +1,4 @@
 class TracksController < ApplicationController
-  include ActionController::Live
-
   before_action :set_track, only: %i[show update destroy audio download merge]
 
   has_scope :by_filter, as: 'filter'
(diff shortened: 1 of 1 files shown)


chvp:
FYI, this was the previous implementation for simple file serving, but this was refactored out in https://github.com/accentor/api/pull/82.

robbevp (approved):

PR #684, opened 2024-11-16
Use ruby that supports jemalloc

chvp:
In production, memory usage of the server can be steadily increased by performing an activestorage request in an infinite loop. The issue can be reproduced in development (though it is less pronounced)

When using a ruby configured with jemalloc, the issue is no longer reproducible in development. So I'm hoping that the same will be true in production.


robbevp (approved):
So maybe more of a fragmentation issue than an actual memory leak? Let's hope that this fixes stuff 👍

PR #721, opened 2025-02-15
Switch to signed token to authenticate users

robbevp:
This PR switched our authentication method from with our device_id and secret to [rails' `generates_token_for`](https://api.rubyonrails.org/classes/ActiveRecord/TokenFor/ClassMethods.html#method-i-generates_token_for).

To avoid clients having to switch immediately, we send the token as if it was the secret and accept the `x-secret` header (or secret param). We also still allow clients to pass the device_id and secret.

## Motivation
My biggest reason for changing is performance. Every api request currently has to go through authentication. In our current implementation, we have to go through bcrypt to check the secret. By switching to `generates_token_for` we only have to check that the signature matches. Considering the amount a requests a user will send to get all albums, artists, tracks, plays, ... from the API - we can get big win if we can speed up each request.

Doing some local request on my dev machine, I can see a big difference for every request. After a few requests of warm up, I get the a ±50% reduction in average request durations. (Note that this wasn't a very scientific test, but just me watching the response times across a bunch of requests for the same controller/action).
* `find_authenticated`: ±100ms
* `find_by_token_for`: ±50ms

We can also see this difference when benchmarking these two methods directly:
```ruby
require 'benchmark'
auth_token = FactoryBot.create(:auth_token)
credentials = { device_id: auth_token.device_id, secret: auth_token.secret }
token = auth_token.generate_token_for(:api)
n = 1000
Benchmark.bmbm do |b|
  b.report("device id + secret") do
    n.times { AuthToken.find_authenticated(credentials) }
  end
  b.report("find_by_token_for") do
    n.times { AuthToken.find_by_token_for(:api, token) }
  end
end
```
```
Rehearsal ------------------------------------------------------
device id + secret  59.289659   0.272631  59.562290 ( 60.044401)
find_by_token_for    0.071776   0.028643   0.100419 (  0.156472)
-------------------------------------------- total: 59.662709sec

                         user     system      total        real
device id + secret  59.029725   0.211262  59.240987 ( 59.538001)
find_by_token_for    0.062914   0.006196   0.069110 (  0.094305)
```

## Format of token
The generated token is of the following format:
```
eyJfcmFpbHMiOnsiZGF0YSI6WzMxLCJmOTM4YjI4YmE0NTdhZDdhMTQ1MGFlYWQiXSwicHVyIjoiQXV0aFRva2VuXG5hcGlcbiJ9fQ==--8635f004299e3933f123665f8e4aadd05362c6a6
```
The base64 encoded json contains the AuthToken's id and device_id.
```
{"_rails"=>{"data"=>[31, "f938b28ba457ad7a1450aead"], "pur"=>"AuthToken\napi\n"}}
``` 

For someone to guess a token they would have to:
* Guess the device_id that matches the AuthToken's id
* Have the correct signature for this message

I ran a manual test with the current web client. When signing in, we still fetch all data and can play audio

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -32,6 +32,7 @@ class ApplicationController < ActionController::API
 
   def authenticate_user
     token = authenticate_with_http_token { |it| AuthToken.find_by_token_for(:api, it) }
+    token ||= AuthToken.find_by_token_for(:api, params[:token])
 
     # While the various clients switch to the authorization header, we also accept the token using the secret header/param
     token ||= AuthToken.find_by_token_for(:api, request.headers[:'x-secret'] || params[:secret])
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,9 +1,10 @@
 class ApplicationController < ActionController::API
   include Pundit::Authorization
+  include ActionController::HttpAuthentication::Token::ControllerMethods
 
   attr_accessor :current_user
 
-  before_action :authenticate_from_token
+  before_action :authenticate_user
   after_action :verify_authorized
   # rubocop:disable Rails/LexicallyScopedActionFilter
   # Most subclasses will have this action, if they don't we also don't need to
@@ -29,13 +30,22 @@ class ApplicationController < ActionController::API
 
   private
 
-  def authenticate_from_token
-    token = AuthToken.find_authenticated(
-      {
-        secret: request.headers[:'x-secret'] || params[:secret],
-        device_id: request.headers[:'x-device-id'] || params[:device_id]
-      }
-    )
+  def authenticate_user
+    token = authenticate_with_http_token { |it| AuthToken.find_by_token_for(:api, it) }
+
+    # While the various clients switch to the authorization header, we also accept the token using the secret header/param
+    token ||= AuthToken.find_by_token_for(:api, request.headers[:'x-secret'] || params[:secret])
+
+    # As a fallback, we still allow users to log in with their device_id and secret
+    if token.nil?
+      token = AuthToken.find_authenticated(
+        {
+          secret: request.headers[:'x-secret'] || params[:secret],
+          device_id: request.headers[:'x-device-id'] || params[:device_id]
+        }
+      )
+    end
+
     self.current_user = token&.user
   end
 
--- a/test/controllers/auth_tokens_controller_test.rb
+++ b/test/controllers/auth_tokens_controller_test.rb
@@ -12,7 +12,26 @@ class AuthTokensControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
-  test 'should create auth_token' do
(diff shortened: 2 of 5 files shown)


robbevp:
@chvp I reworked this after our discussion yesterday. We should be able to roll this out without any client needing to change their implementation immediately.

I split this in two commits, since I wasn't sure whether we wanted to remove the secret immediately and force the user to sign in again. I think we can just bite the bullet now, but I'm okay with leaving out this commit for now.

test/controllers/auth_tokens_controller_test.rb
```diff
@@ -12,7 +12,15 @@ class AuthTokensControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
-  test 'should create auth_token' do
+  test 'should allow user to authorize using the fallback header' do
```
robbevp, line 15: I only added this fallback test in one controller. I don't think we need to test this more than once.

test/test_helper.rb
```diff
@@ -62,46 +62,43 @@ class ActiveSupport::TestCase
...
-    self.credentials = {}
+    self.api_token = nil
   end
 
-  def merge_args(args)
-    args ||= {}
-    args[:headers] = args[:headers] || {}
-    creds = credentials || {}
-    { as: :json }.merge(args || {}).merge(headers: creds.merge(args[:headers]))
+  def merge_args(**args)
```
robbevp, line 75: I had the feeling we were doing a lot of unnecessary checks here, so I simplified this quite a bit

chvp (commented):
Nice, looks good to me!

I would keep that second commit for later though. Once all clients have been migrated, we can force a re-login and remove the fallback code in the `application_controller` and the `auth_tokens_controller` at the same time.

robbevp:
Removed the second commit (I'll open a draft PR with that commit)

robbevp:
@chvp Any chance you could take another look at this? This should be ready to deploy without disrupting any users/sessions or breaking any clients that haven't updated yet (both for existing sessions and session with the new tokens)

app/controllers/application_controller.rb
```diff
@@ -29,13 +30,23 @@ def add_pagination_headers(collection)
...
-  def authenticate_from_token
-    token = AuthToken.find_authenticated(
-      {
-        secret: request.headers[:'x-secret'] || params[:secret],
-        device_id: request.headers[:'x-device-id'] || params[:device_id]
-      }
-    )
+  def authenticate_user
+    token = authenticate_with_http_token { |it| AuthToken.find_by_token_for(:api, it) }
+    token ||= AuthToken.find_by_token_for(:api, params[:token])
```
chvp, line 35: This is for things like audio requests, I assume?

chvp (approved):
Nice! Approved, but I disabled auto-merge first because of the question.

app/controllers/application_controller.rb
robbevp: Yes, streaming audio is probably the only case where we cannot (always) use headers

PR #723, opened 2025-02-17
Remove secret from auth tokens

robbevp:
Followup on #721 that removes the fallback we added there


robbevp:
This is ready for review - but we need to wait with merging/deploying until all clients have had the chance to switch to the new tokens

db/migrate/20250216134558_remove_hash_secret_from_auth_tokens.rb
```diff
@@ -0,0 +1,17 @@
...
+    add_column :auth_tokens, :hashed_secret, :string, null: true
+
+    AuthToken.find_each do |at|
+      at.secret = SecureRandom.urlsafe_base64(48)
+      at.hashed_secret = BCrypt::Password.create(secret, cost: Rails.configuration.token_hash_rounds)
+      at.save!
+    end
+
+    change_column_null :auth_tokens, :hashed_secret, false
+  end
```
chvp, line 16: Not sure about this `down` migration. The hashed_secret will not work. I think it would be more honest to just delete all the existing auth tokens.

db/seeds.rb
```diff
@@ -2,7 +2,7 @@
 
 if Rails.env.development?
   token = AuthToken.create(user: user, user_agent: 'Rails')
-  token.update(hashed_secret: BCrypt::Password.create('secret', cost: 1), device_id: 'device-id')
+  token.update(device_id: 'device-id')
 end
```
chvp, line 6: Does it make sense to still add this? IIUC there is no way to use this token right?

db/migrate/20250216134558_remove_hash_secret_from_auth_tokens.rb
chvp: (Also, the configuration for `token_hash_rounds` should probably be removed, since it isn't used in the actual application anymore.)

robbevp: I do think we should remove the configuration, and just hardcode the cost here

db/seeds.rb
robbevp: It's pretty easy to use this token by getting the following in a rails console: `AuthToken.first.generate_token_for(:api)`. I tend to use this when trying things in the api, so that I can make request using curl rather having to go through the web client.
(I mean, it isn't much more work to do something like `FactoryBot.create(:auth_token, user: User.first).generate_token_for(:api)`)

robbevp:
iOS/macOS was updated here: https://github.com/accentor/ios_macos/pull/74
Game was updated here: https://github.com/accentor/game/pull/109

db/migrate/20250216134558_remove_hash_secret_from_auth_tokens.rb
chvp: Sure.

db/seeds.rb
```diff
@@ -2,7 +2,7 @@
 
 if Rails.env.development?
   token = AuthToken.create(user: user, user_agent: 'Rails')
-  token.update(hashed_secret: BCrypt::Password.create('secret', cost: 1), device_id: 'device-id')
+  token.update(device_id: 'device-id')
```
chvp, line 5: This could just be `AuthToken.create(user:, user_agent: 'Rails', device_id: 'device-id')` not. I think it was just separated because of the `before_validation` in AuthToken that generated the secret. (Which wasn't even a problem now that I look at it.)

chvp (approved):

PR #813, opened 2025-09-14
Filter token from being logged

chvp:
Noticed while looking at the logs for a change to the android app that the token was in there in clear text, we should probably avoid that :grimacing:


robbevp (approved):
👍

PR #970, opened 2026-07-31
Use flake's devshell in CI

robbevp:
This PR updates the CI to use the flake's devshell rather than setting up our dependencies from scratch. By doing this, we ensure that we're using the exact same dependencies and setup in the CI as we do in development (and if production if this uses nixos)

Based on my experience with other projects, I would expect this to be a bit slower for the lint action and faster for the tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,6 @@ jobs:
         with:
           name: accentor
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
-          skipPush: true
       - name: "Setup database"
         run: |
           nix develop --command pg:start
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -14,47 +14,44 @@ jobs:
...
+      - uses: actions/checkout@v7
+      - uses: cachix/install-nix-action@v31
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Cachix
+        uses: cachix/cachix-action@v17
+        with:
+          name: accentor
+          authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
+          skipPush: true
```
robbevp: I would consider not adding the `skipPush` option. This will push a bit more to cachix, but could speed up the CI if there are dependencies that aren't in cache.nixos.org and that haven't been pushed by our cachix workflow

chvp: Sure

chvp (approved):

PR #976, opened 2026-08-01
Fix simplecov deprecation warnings

chvp:
Noticed this recently when looking at the test output.


robbevp (approved):
👍  (sorry - seemed to have missed the review request)

PR #985, opened 2026-08-15
Modernize the release workflow

chvp:
Similar to https://github.com/accentor/api-client-js/pull/840, but without the npm publish part.


robbevp (approved):
👍

PR #986, opened 2026-08-16
Do a policy check at the resource level before any other callbacks

chvp:
Expect for the auth callback, of course. This fixes a leak in our policy/resource handling, where a user could know whether a record with a certain ID existed based on the HTTP response.

A similar issue existed for logged-in users for other user's secret records (auth tokens and playlists, in practice). By policy scoping the `find` calls, authorized users will always get a 404 for records they don't have access to, whether that record exists or not.

- [x] I've added tests relevant to my changes. (Well, I've mostly updated the existing tests, to be more clear.)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -62,10 +62,10 @@ class ApplicationController < ActionController::API
 
   def user_not_authorized(exc)
     status = current_user.present? ? :forbidden : :unauthorized
-    render json: { errors: [{ model: exc.policy.record.model_name.singular, type: status, action: exc.query }] }, status:
+    render json: { errors: [{ model: exc.record.model_name.singular, type: status, action: action_name }] }, status:
   end
 
   def model_not_found(exc)
-    render json: { errors: [{ model: exc.model.downcase, type: :not_found }] }, status: :not_found
+    render json: { errors: [{ model: exc.model.underscore, type: :not_found }] }, status: :not_found
   end
 end
--- a/app/controllers/auth_tokens_controller.rb
+++ b/app/controllers/auth_tokens_controller.rb
@@ -15,7 +15,8 @@ class AuthTokensController < ApplicationController
   end
 
   def create
-    authorize AuthToken
+    # We don't use pundit in this action since it's configured to require a user
+    skip_authorization
 
     user = User.find_by(name: params[:name])
     unless user.try(:authenticate, params[:password])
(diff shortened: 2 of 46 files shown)


chvp:
The coverage going down is because we don't check the `scope.none` case for a lot of the policies. We should probably introduce unit tests for our policies, but let's not do that in this PR (IMO).

app/controllers/auth_tokens_controller.rb
```diff
@@ -42,8 +42,14 @@ def destroy
 
   private
 
+  def check_resource_access
+    return true if action_name == 'create'
```
chvp: This could also be a `skip_before_action :check_resource_access, only: :create`, but I think I prefer this?

app/controllers/application_controller.rb
```diff
@@ -60,12 +61,14 @@ def authenticate_user
...
+  def check_resource_access = raise NotImplementedError, 'should be implemented by subclass'
+
   def user_not_authorized(exc)
     status = current_user.present? ? :forbidden : :unauthorized
     render json: { errors: [{ model: exc.policy.record.model_name.singular, type: status, action: exc.query }] }, status:
   end
 
   def model_not_found(exc)
-    render json: { errors: [{ model: exc.model.downcase, type: :not_found }] }, status: :not_found
+    render json: { errors: [{ model: exc.model.underscore, type: :not_found }] }, status: :not_found
```
chvp: Changed this because it rendered as `authtoken` for the auth token controller, which seems wrong.

robbevp (commented):
I'm not sure about this one, using `check_resource_access` feels a bit weird to me:
* I find it more natural if the public methods of policies correspond to the actions in the controller
* I think it solves the same problem as you do with `policy_scope`: I don't see a lot of cases where the user would get passed `policy_scope` (and the policy that matches the action) but would currently get stopped by `check_resource_access`

Maybe we should split this problem in two:
1. A user who is not logged in, should have no way to find out details about what's in the database
2. A user who is logged in, should not know about the ids of objects it can't see.

For the first problem, I think we might be better of with an early return of unauthorized. Rather than having to check this in each policy, we know that signing in is the only action that doesn't require a user. I guess either by:
* Updating our `authenticate_user` to require the user to be present or returning unauthorized (and then using `skip_before_action` for the sign in action)
* Adding something like `require_user!` and adding this as a before action (either per controller, or with `skip_before_action`).
* We could make our policies [a closed system](https://github.com/varvet/pundit#closed-systems), and also make some exception for creating auth tokens

For the second problem (which I don't worry about as much, I'm not sure what I could do by knowing the id of a secret playlist someone else made 🤷 ), the policy_scope should be enough?

robbevp (changes requested):
👍 Just noticed one method that should be removed

robbevp (approved):
👍

PR #988, opened 2026-08-20
Rework our nix setup

chvp:
Similar to https://github.com/accentor/web/pull/1527.

This PR does the following:
- Gets rid of our dependency on flake-utils. It's not really necessary in modern nix flakes (see also https://nixcademy.com/posts/new-default-nix-flake-template/).
- Moves the package and shell definition to separate files
- Changes the way we filter files in the package source definition (using lib.fileset and an allow list instead of a deny list)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,26 @@
+{ stdenv
+, lib
+, accentor-api-env
+}:
+
+let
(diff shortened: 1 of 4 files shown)


robbevp (approved):
👍 

(Though I think you accidentally linted the gemset.nix - there's some merge conflicts in that file)

PR #989, opened 2026-08-21
Add tests for our policies

chvp:
The only changes to our actual app code is some hardening in the permitted attributes. This shouldn't result in any behaviour changes.

### DIFF
--- a/.github/workflows/cachix.yml
+++ b/.github/workflows/cachix.yml
@@ -17,11 +17,8 @@ jobs:
         with:
           name: accentor
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
-          skipPush: true
       - run: nix build -L --no-link .#accentor-api
       - run: nix build -L --no-link .#accentor-api.env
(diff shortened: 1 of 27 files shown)


robbevp (commented):
All for hardening and testing the policies! I'll keep this open, until we resolve the PR earlier in the stack.

A few open suggestions:
* [There's a gem](https://rubygems.org/gems/pundit_assertions) (that I wrote) that provides some custom assertions for policies, so we can just write `assert_permitted ...`. Though I also understand not wanting to add a gem for rather trivial things like this
* I would avoid the confusion of `anonymous user` and use something like `guest` (since an anonymous user is in fact the absence of a user)
* I have the habit of checking all user types for each action/scope/method, just for clarity. So often do something like:
  ```rb
  test 'should allow index for users and moderators' do
    assert_predicate Pundit.policy!(@user, Album), :index?
    assert_predicate Pundit.policy!(@moderator, Album), :index?
    assert_not_predicate Pundit.policy!(nil, Album), :index?

robbevp (changes requested):
👍 I'd just split off the changes in the workflows and flake to keep the PR focused on one thing

robbevp (approved):
👍

PR #994, opened 2026-08-23
Make dependencies job more lightweight

chvp:
Because the dependencies job didn't use cachix and used the ruby from the gems, it needed to build Ruby with jemalloc support, which took a long time. This changes the dependencies job to use cachix, and makes sure the deps shell doesn't go through the gemset derivation.


robbevp (approved):
👍

